PR #1, opened 2016-05-22
Prevent fatal error in autoloader

nohatssir:
Switches from require_once to include_once. The former throws a fatal error when a file isn't found while the later just emits a warning. Because of the fatal error autoloaders registered after the one provided by OMV won't be executed. The error message is also not quite obvious because of this, since it tells you which file that failed to load instead of which class.

### DIFF
--- a/deb/openmediavault/usr/share/php/openmediavault/autoloader.inc
+++ b/deb/openmediavault/usr/share/php/openmediavault/autoloader.inc
@@ -27,5 +27,5 @@ spl_autoload_register(function($className) {
 	array_unshift($parts, "openmediavault");
 	$pathName = strtolower(sprintf("%s.inc", implode(DIRECTORY_SEPARATOR,
 	  $parts)));
-	require_once $pathName;
+	include_once $pathName;
 });


openmediavault:
I don't think this is a good idea. If some files are not available, an error should be thrown because only a warning may result in an incorrect working software which might produce other troubles. The system itself must be in a valid state to work correct. If this is not the case, a error is the correct answer.

nohatssir:
Yes, but that is kind of wrong. Autoloading doesn't fail if an autoloader doesn't find a file, it fails if a class is not loaded by the time all autoloaders has tried to load the class.

By throwing an error in the autoloader other autoloaders can't try to autoload the class. If the class isn't found after all autoloaders have tried to load the class an exception will be thrown anyway but with a much clearer message, `PHP Fatal error:  Uncaught Error: Class 'OMV\Class' not found in /path/to/file:line`. The state will still be correct and this makes it so that you directly see which file it is that tries to use the class and that makes it easier to trace the error. E.g. is it a typo?, have I forgot to include the namespace?, etc.

As it is now, you can't trace down where the error originates from. Let's say that you forget to include the full namespace or an use statement in OMV and just calls `new SystemCtl()` by mistake. The error you will get is this:
```
PHP Fatal error:  require_once(): Failed opening required 'openmediavault/engined/rpc/systemctl.inc' (include_path='.:/usr/share/php:/usr/share/pear') in /usr/share/php/openmediavault/autoloader.inc on line 30
PHP Stack trace:
PHP   1. {main}() /usr/sbin/omv-engined:0
PHP   2. require_once() /usr/sbin/omv-engined:344
PHP   3. spl_autoload_call() /usr/sbin/omv-engined:29
PHP   4. {closure:/usr/share/php/openmediavault/autoloader.inc:22-31}() /usr/sbin/omv-engined:0
```
This error doesn't help you and makes it guesswork to trace the origin and that's what the patch resolves.

PR #3, opened 2016-06-19
Update Jobs.js

subzero79:
Display the corresponding device for the job entry


subzero79:
At the moment the "External Storage Device" column is not displaying the value. I didn't want to reopen the BT thread again, is a minimal change. Stripping down the "/dev/disk/by-id" part string is a little bit beyond my skills. So is just a reminder to correct and also in case you want to "prettify" the value.

votdev:
Thanks

PR #9, opened 2016-11-27
Fix avoid mail notification every week about smart db update

lucagiove:
log instead of echo to stdout
removed exit to return the eval return code


openmediavault:
log instead of echo to stdout removed exit to return the eval return code

PR #10, opened 2016-12-20
Mantis 0001645: Removed output of mkpatidx from APT::Update conf

lucagiove:
Because of mkaptidx the /etc/cron.daily/apt script has standard output
that cause a daily mail notification.

Related issues:
http://bugtracker.openmediavault.org/view.php?id=1645
http://bugtracker.openmediavault.org/view.php?id=1334

### DIFF
--- a/deb/openmediavault/etc/apt/apt.conf.d/99openmediavault-mkaptidx
+++ b/deb/openmediavault/etc/apt/apt.conf.d/99openmediavault-mkaptidx
@@ -1,3 +1,3 @@
 // Create the plugin index.
-APT::Update::Post-Invoke-Success { "which omv-mkaptidx >/dev/null 2>&1 || exit 0; omv-mkaptidx || true"; };
-DPkg::Post-Invoke { "which omv-mkaptidx >/dev/null 2>&1 || exit 0; omv-mkaptidx || true"; };
+APT::Update::Post-Invoke-Success { "which omv-mkaptidx >/dev/null 2>&1 || exit 0; omv-mkaptidx >/dev/null || true"; };
+DPkg::Post-Invoke { "which omv-mkaptidx >/dev/null 2>&1 || exit 0; omv-mkaptidx >/dev/null || true"; };


lucagiove:
I left stderr so that error can be reported via mail instead

votdev:
The cron-apt job is executed daily, so you should always get an email notification, regardless if omv-mkaptidx prints output or not.

lucagiove:
Yes but this mail is not generated by cron-apt but from _/etc/cron.daily/apt_ 
At line 439:
`if eval apt-get $XAPTOPT -y update $XSTDERR; then`
Executes apt-get update that executes mkaptidx in APT::Update::Post-Invoke-Success

votdev:
With your patch no output is done in any case, e.g. when you do a 'apt-get update' in the console. This is not wanted, omv-mkaptidx should tell the user what he is doing.

What is the output of 'apt-config dump | grep APT::Periodic::Verbose' in your system. By default every output from stdout and stderr should be redirected to /dev/null with VERBOSE=0 (default), see line 355.

votdev:
Due the fact that there is no other option to prevent this output i have accepted this PR. But i think this is a bug in apt-get because it should take care that, if option -q is set, the output of all child processes (e.g. Post-Invoke scripts) is suppressed.

lucagiove:
Maybe we could read the verbosity and set accordingly the redirection like in the /etc/cron.d/apt script itself but not sure whether it's worth...

```
# Set VERBOSE mode from  apt-config (or inherit from environment)
VERBOSE=0
eval $(apt-config shell VERBOSE APT::Periodic::Verbose)
debug_echo "verbose level $VERBOSE"
if [ "$VERBOSE" -le 2 ]; then
    # quiet for 0,1,2
    XSTDOUT=">/dev/null"
    XSTDERR="2>/dev/null"
    XAPTOPT="-qq"
    XUUPOPT=""
else
    XSTDOUT=""
    XSTDERR=""
    XAPTOPT=""
    XUUPOPT="-d"
fi
if [ "$VERBOSE" -ge 3 ]; then
    # trace output
    set -x
fi
```

votdev:
Too much hassle for such a simple issue. And APT::Periodic::Verbose does not have anything to do with omv-mkaptidx.

PR #11, opened 2017-01-20
proFTPd settings: Changing from old TLSv1.0 to the more secure TLSv1.2

sfu420:


andruare:
Why is omv still using ProFTPD 1.3.4a?

Current Release: 1.3.5d
http://www.proftpd.org/docs/RELEASE_NOTES-1.3.5d

openmediavault:
OMV does not ship packages related to the operating system. It uses the packages shipped and maintained by the Debian project (e.g. https://packages.debian.org/wheezy/proftpd-basic). If you want to update you have to compile it yourself.

PR #14, opened 2017-03-14
RPC/diskmgmt/rescan detect disk size change as well

ZsZs73:
The existing script in the rescan function scans the SCSI bus for new/removed devices only.
The new part of the function will rescan every scsi device to detect size changes.
This will result in following kernel message:
[86617.420189] sd 2:0:1:0: [sdb] 31457280 512-byte logical blocks: (16.1 GB/15.0 GiB)
[86617.420219] sdb: detected capacity change from 15032385536 to 16106127360

### DIFF
--- a/deb/openmediavault/usr/share/openmediavault/engined/rpc/diskmgmt.inc
+++ b/deb/openmediavault/usr/share/openmediavault/engined/rpc/diskmgmt.inc
@@ -285,5 +285,11 @@ class OMVRpcServiceDiskMgmt extends \OMV\Rpc\ServiceAbstract {
 		  "\$(find /sys/class/scsi_host -iname \"host*\" -type l); ".
 		  "do echo \"- - -\" | tee \${hostdir}/scan >/dev/null; done");
 		$script->execute();
+		// Execute shell script to detect SCSI device size changes.
+		$script = new \OMV\System\ShellScript(
+		  "[ -x /sys/class/scsi_device ] && for devicedir in ".
+		  "\$(find /sys/class/scsi_device -iname \"*:*\" -type l); ".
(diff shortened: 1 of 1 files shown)


ryecoaaron:
hdparm -z /dev/sdX works in this situation as an alternative to echoing 1 to rescan.  Would eliminate the pipe and tee.

votdev:
Is the use of sysfs not better than using an external tool?

ryecoaaron:
Piping to tee is using an external program.  I wasn't saying the original solution is bad.  Just suggesting an alternative.

ZsZs73:
According to 'man hdparm':

> -z
> Force a kernel re-read of the partition table of the specified device(s).

However I tried and it indeed re-checks the disk size as well. I do not even have a partition on that disk as LVM uses it for a PV.
I would still use sysfs.

Regarding the 'tee' in the pipe topic. I just wanted to keep the original approach - coming from the hosts scan section above.
But I agree with Aaron and I also do not see the point of using the tee command in the pipe and then redirect the pipe to /dev/null.

In case there is no other justification of using 'tee' here, 
`"do echo 1 | tee \${devicedir}/device/rescan >/dev/null; done");`
can be replaced with
`"do echo 1 > \${devicedir}/device/rescan; done");`
I tested it and it works well. 
The HBA scan also works without tee. Should I modify the pull request according to this?

Actually in OMV 3.0.65 with stock 3.16.0-4 kernel adding/removing a disk shows up in the physical disk list by simply clicking the "Physical Disks".
Regardless of this I would keep the HBA scan section as well.

votdev:
The only reason for using the tee command is that in OMV 1.x it was necessary that no command has an output, otherwise this output was send as part of the RPC response. Nowadays it is not necessary to take care about this.

PR #18, opened 2017-05-13
Change padding from 6/7px to 5px

Skaronator:
Its reduce the size of white space between the menu buttons. You'll see more buttons and have less scroll while looking still good.

Here a example:
image

### DIFF
--- a/deb/openmediavault/var/www/openmediavault/css/theme-triton.css
+++ b/deb/openmediavault/var/www/openmediavault/css/theme-triton.css
@@ -150,6 +150,10 @@ body > .x-mask {
 .x-grid-tree-node-expanded .x-tree-elbow-end-plus {
 	color: #000000;
 }
+.x-tree-icon, .x-tree-elbow-img, .x-tree-checkbox {
+	margin-top: -5px;
+	margin-bottom: -5px;
+}
 /******************************************************************************/
 /* Grid */
 .x-grid-item-focused .x-grid-cell-inner:before {
@@ -162,9 +166,9 @@ body > .x-mask {
 	opacity: 0.7;
 }
 .x-grid-cell-inner {
-	padding: 4px;
+	padding-top: 5px;
+	padding-bottom: 5px;
 }
-
 /******************************************************************************/
 /* Workspace node tree */
 .x-workspace-node-tree .x-panel-header-default {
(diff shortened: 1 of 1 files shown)


votdev:
I suggest to use

`.x-grid-cell-inner {
    padding-top: 5px;
    padding-bottom: 5px;
}`

because 4px are too small in correlation to the top and bottom toolbars of a grid.

Skaronator:
Even 3px are fine IMO but going to change it to 5px.

One question: What is the purpose of `<div class=" x-tree-elbow-img x-tree-elbow-line" role="presentation"></div>`

`.x-tree-icon, .x-tree-elbow-img, .x-tree-checkbox` those classes have the negative margin of `x-grid-cell-inner` padding.

Skaronator:
Updated. A Preprocessors like SASS or LESS might be very helpful to keep consistency over various elements.

votdev:
Indeed it would be great to use SASS or LESS, but this is not possible (as far as i know) without buying the whole Sencha ExtJS package because this contains the necessary tool chain and files. But i am open for contributions in this behaviour.

Skaronator:
A switch to ReactJS might be good but you won't have the time to do this task.

votdev:
It is not possible nor wanted to switch the UI framework because the code base is to big now. Also ExtJS is one of the best framework, better than AngularJS for example which i am working with, too.

votdev:
I have to decline this PR after testing because:

- the layout of the tree cells is totally shifted and asymetric.
- the whole UI layout is not fit anymore.

Please use /var/www/openmediavault/css/theme-custom.css to customize the UI to your needs.

Skaronator:
Weird. Something is wrong with the negative margin. Still don't know why it is there in first place. I've removed it and now its working just fine.

Could you verify that its working now?

Simply put this in the Dev Console when you're on the OMV Interface.

```
var css = document.createElement("style");
css.type = "text/css";
css.innerHTML = " .x-grid-cell-inner, .x-grid-cell-inner-treecolumn { padding-top: 5px; padding-bottom: 5px;}";
document.head.appendChild(css);
```

I've updated this in my fork as well but the PR isn't updating: https://github.com/Skaronator/openmediavault/commit/2cbd3772da9b2adc643ff24134f170b8eedfe4b7

votdev:
Fixed with https://github.com/openmediavault/openmediavault/commit/5e6b69f0851e8fab6ce41a234c7a8da67d63b6a2.

Skaronator:
Thanks!

PR #19, opened 2017-05-16
Update rsync test if source dir is mounted

neeqkowlah:
add test if source directory is mount (in case of source external drive for exemple)


votdev:
I have to decline this PR because of the following reasons:

- the target branch is incorrect, it must be 3.x.
- in pull mode the variable $srcmntdir is not used and will produce an error.

votdev:
Fixed with https://github.com/openmediavault/openmediavault/commit/f877984e02571e5bdb317819e7d8d550eed88144.

PR #20, opened 2017-05-22
3.x - navigation menu missing from webUI

EagleEye2004:
As discused in the following topic, the tree panel should be available on touch devices and tablets. 
http://forum.openmediavault.org/index.php/Thread/18323-navigation-menu-missing-from-webUI/

Since ext js tablet also include smartphones running firefox only desktop and touch devices added.

Tested:
- Edge, Chrome on Surface Pro 3 (Firefox is not testable since language is not selectable in login screen)
- Edge, Chrome on Desktop (without Touch)
- Chrome, Firefox on Nexus 6 (Android 7.1.1)


EagleEye2004:
I have selected the wrong branch

PR #22, opened 2017-05-22
Switches must be interactive

R0maNNN:
Led must indicate.
Toggle button must be interactive.


votdev:
Thanks for your contribution. I will not merge it right now because in my opinion the WebUI does not look as good as before. Your PR is correct because the switches indicate an interactive element which it is not, but the changed look isn't that good. I will think about what can be done to improve the look.

votdev:
See http://bugtracker.openmediavault.org/view.php?id=1745.

R0maNNN:
I think the true way is to make switches interactive. In case of impossibility of turning on the component without configuring, it must be gray.

switched on -> image

switched off -> image

switched off (need configuring) -> image

It looks like:
default

R0maNNN:
@votdev 
please look at

votdev:
Done by https://github.com/openmediavault/openmediavault/commit/81bbf607725b3a269e54fa4b2970f8f54a88a9e7.

PR #23, opened 2017-05-24
Clamav fails to start on new OMV installation due to the missing /var/run/clamav path

guarnacciaa:
On a new OMV installation Clamav fails to start with the following errors: 

Wed May 24 13:33:31 2017 -> ERROR: Can't save PID to file /var/run/clamav/freshclam.pid: No such file or directory
Wed May 24 13:34:15 2017 -> WARNING: Clamd was NOT notified: Can't connect to clamd through /var/run/clamav/clamd.ctl: No such file or directory


votdev:
This is a bug in the clamav package in my opinion. If we should include this as a workaround to fix the package bug, then i would like to see this in https://github.com/openmediavault/openmediavault/blob/3.x/deb/openmediavault-clamav/debian/openmediavault-clamav.postinst instead of the omv-mkconf script.

Please also create the PR against the 3.x branch and not master.

votdev:
Issue has been fixed with https://github.com/openmediavault/openmediavault/commit/4b55bc9011d084820f9982175a5837309ba36ca6.

PR #26, opened 2017-07-17
Workaround for memory leak when calling OMV.MessageBox.guru

emontnemery:
Workaround for memory leak when OMV.MessageBox.guru is called repeatedly from Rpc.js

### DIFF
--- a/deb/openmediavault/var/www/openmediavault/js/omv/Rpc.js
+++ b/deb/openmediavault/var/www/openmediavault/js/omv/Rpc.js
@@ -30,6 +30,7 @@ Ext.define("OMV.Rpc", {
 	requires: [ "OMV.window.MessageBox" ],
 	singleton: true,
 	autoAbort: false,
+	guru: null,
 
 	config: {
 		disabled: false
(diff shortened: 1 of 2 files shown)


deb/openmediavault/var/www/openmediavault/js/omv/Rpc.js
```diff
@@ -167,7 +168,11 @@ Ext.define("OMV.Rpc", {
 				me.setDisabled(true);
 				// Display a dialog forcing user to click 'OK' to reload
 				// the page.
-				OMV.MessageBox.guru({
+				if (this.guru)
+				{
+					this.guru.destroy();
+				}
+				this.guru = OMV.MessageBox.guru({
```
votdev: I'm wondering why you need to take care to destroy the object? The window is created once, then the code path is never called, see me.setDisabled(true). So there is only one instance until the user clicks the mouse button. After that the whole page is reloaded.

emontnemery: The problem is when the user session expires but there is a recurring query:
The recurring query is still active, and a new guru meditation window will be created each time there is an (unsuccessful) reply to the query.

votdev:
The patch does not fix the issue at all. This is done by https://github.com/openmediavault/openmediavault/commit/93da4bbf708a252515f9de5bf05bb4b2ffa263de.

emontnemery:
Sure, it was more meant as a workaround / to show the problem.
Thanks for a fantastic product!

PR #29, opened 2017-07-20
Fixed ticket #1772: SMART overview shows red, details are all green

modelrockettier:
When OpenMediaVault gets a drive's overall SMART status, it checks the normal attributes, but then also checks the Reallocated_Sector_Ct (5) and Current_Pending_Sector (197) attributes.

Then when viewing a drive's detailed SMART attributes, these same checks are not performed. If you have a drive with at least 1 reallocated or pending sector, this will cause the drive's overall status to be red, but all the detailed SMART attributes will be green or grey.

To make the status consistent, I added checks to the SMART prefailure attribute checks, which will now cause those 2 attributes to show up as red when their raw values are > 0.

### DIFF
--- a/deb/openmediavault/usr/share/php/openmediavault/system/storage/smartinformation.inc
+++ b/deb/openmediavault/usr/share/php/openmediavault/system/storage/smartinformation.inc
@@ -252,7 +252,7 @@ class SmartInformation {
 			$attrData['assessment'] = self::SMART_ASSESSMENT_BAD_ATTRIBUTE_IN_THE_PAST;
 			return;
 		}
-		// mark bad sector attributes as red when rawvalue >= 1, ticket #1772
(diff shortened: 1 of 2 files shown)


deb/openmediavault/usr/share/php/openmediavault/system/storage/smartinformation.inc
```diff
@@ -252,6 +252,13 @@ class SmartInformation {
 			$attrData['assessment'] = self::SMART_ASSESSMENT_BAD_ATTRIBUTE_IN_THE_PAST;
 			return;
 		}
+		// mark bad sector attributes as red when rawvalue >= 1, ticket #1772
```
votdev: Please use this comment.

// Check for bad sectors (Reallocated_Sector_Ct and Current_Pending_Sector).

PR #30, opened 2017-07-20
Fix sysinfo report when /etc/apt/sources.list is missing

stweil:
Signed-off-by: Stefan Weil <sw@weilnetz.de>


stweil:
The new code with `[ ... ]` instead of `test` is fine for me. I suggest to keep the caption and the separator for `sources.list` to indicate that this part was checked. There is also always a caption and a separator for `sources.list.d`, no matter whether there are files in that directory or not. Feel free to tell me any more required changes (or simply add them) before you merge that change.

PR #37, opened 2017-09-11
btrfs-progs >= 4.8.4 btrfs-show-super depricated

darkman1983:
The btrfs-show-super is merged into the main btrfs: btrfs inspect dump-super

### DIFF
--- a/deb/openmediavault/usr/share/php/openmediavault/system/filesystem/btrfs.inc
+++ b/deb/openmediavault/usr/share/php/openmediavault/system/filesystem/btrfs.inc
@@ -209,7 +209,7 @@ class Btrfs extends Filesystem {
 
 		$cmdArgs = [];
 		$cmdArgs[] = escapeshellarg($this->getDeviceFile());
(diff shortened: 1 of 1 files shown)


darkman1983:
System is: Debian 9.1
Kernel: 4.12 (own compile)
Btrfs-Progs: 4.13
OMV: 4.0.4

Maybe there should be a btrfs version constant in the PHP file? and than compare version to select which command to use.

votdev:
OMV uses Debian Stretch [btrfs-progs](https://packages.debian.org/stretch/btrfs-progs) 4.7.3, so this patch will break the code. OMV will not support other versions than the packages delivered through the Debian repositories. Backports are also not supported. This is to keep the code base clean and do not have to check which version is running.

votdev:
I've added a method to unregister a filesystem backend, see [here](https://github.com/openmediavault/openmediavault/commit/d6a16be9cec64430b4d30bd7340f3a12047d78b5). With this feature you should be able to implement a plugin that ships an adapted BTRFS filesystem backend for the latest kernel and userland.

PR #38, opened 2017-09-17
Update 20kernel_backports

ryecoaaron:


ryecoaaron:
I just realized that lsb_release (lsb-release package) might not be on every system.  Not sure if you want to add it as a dependency or just go back to a hard-coded release name.

PR #41, opened 2017-10-08
Add the capability to customize bind options for NFS exports via envi…

subzero79:
…ronment variable. Main usage is to create fstab entries when is required to use noauto


subzero79:
An example of this is for using encrypted drives. A custom setup with a special systemdl target that holds all startup services (except ssh) to wait for drive unlocking. If noauto is not present local-fs.target will fail as attempting to mount drives or binds. Binds can be related to encrypted block devices, dropping then to emergency or rescue target. An example of this is show here

https://blog.iwakd.de/headless-luks-decryption-via-ssh

PR #42, opened 2017-10-21
typo

sprtz:


votdev:
Thanks for the fix, i've modified the text at translate.openmediavault.org.

PR #46, opened 2017-11-20
Stop and disable the mount bind unit.

subzero79:
Disabling this will remove the symlink from the install target


votdev:
Good idea, i had that also in mention, too.

subzero79:
I though now had to be passed to be disabled and stopped.

PR #55, opened 2018-02-26
Fix cron script of recyclebin delete

Fantu:
On server with openmediavault package 3.0.96 I found that never clean the recycle of all shares even if setted correctly (causing big problems of full disks).
After have investigated I found that /var/lib/openmediavault/cron.d/samba-recycle-* file executed manually always exit with status 0 without delete the files that must delete (even if find commands work correctly), exit with both lock dir exist and not.
This PR solves this problem (simpler and faster solution that came to my mind, already tested)
When will be applied can be backported also to 3.0.x please?

### DIFF
--- a/deb/openmediavault/usr/share/openmediavault/mkconf/samba.d/40recyclebin
+++ b/deb/openmediavault/usr/share/openmediavault/mkconf/samba.d/40recyclebin
@@ -69,9 +69,9 @@ done
 # Exit if 'recyclemaxage' is zero.
 [ "\${amin}" -eq "0" ] && exit 0
 # Exit if another job is running.
-if ! mkdir "${lockfile}" &>/dev/null; then
-	exit 0
-fi
+[ -e ${lockfile} ] && exit 0
+# Create lock
+mkdir "${lockfile}" &>/dev/null
 # Initialize the trap to cleanup on exit.
 trap "rm -rf ${lockfile}; exit" 0 1 2 5 15
 # Exit if the recycle bin directory does not exist.


deb/openmediavault/usr/share/openmediavault/mkconf/samba.d/40recyclebin
```diff
@@ -69,9 +69,9 @@ done
 # Exit if 'recyclemaxage' is zero.
 [ "\${amin}" -eq "0" ] && exit 0
 # Exit if another job is running.
-if ! mkdir "${lockfile}" &>/dev/null; then
-	exit 0
-fi
+[ -e ${lockfile} ] && exit 0
```
votdev: The initial solution has been choosen because mkdir is atomic. With your implementation the creation and checking of the lock file is not atomic anymore because it is splitted into separate operations.

votdev: The problem should be solved if the trap is relocated before the mkdir command. I have to check that.

Fantu: With actual solution script never do its work of cleaning recycle because exit always with exit status 0 before, first time creating "lockfile" and after because exist, move trap before delete the "lockfile" on exit but still exit without execute the clean, the only working solution is split create and check "lockfile" (I not know way to make it working without split)

votdev:
This pull request is obsolete now. See https://github.com/openmediavault/openmediavault/commit/d903f254c72e7605362b5f1395e89e674b11eb61.

PR #58, opened 2018-03-03
Fix http request timeout ignored

CyBHFal:
Ext JS 6 does not use timeout member variable but _timeout due to configPrefixed=true. In fact, it reads the value with the getter getTimeout() which choose timeout or _timeout depending on $configPrefixed=true.
Instead of setting timeout member variable we have to use setter setTimeout() to initialize the right member variable.

To increase default http request timeout, edit following configuration files:
-/etc/openmediavault/php.ini: increase max_execution_time (in seconds)
-/etc/nginx/sites-enabled/openmediavault-webgui: increase fastcgi_read_timeout (in seconds)
-/etc/default/openmediavault: set OMV_HTTPREQUEST_TIMEOUT (in milliseconds)

I set OMV_HTTPREQUEST_TIMEOUT to the exactly wanted timeout and the two others with additionnal 15s

### DIFF
--- a/deb/openmediavault/var/www/openmediavault/js/omv/data/proxy/Rpc.js
+++ b/deb/openmediavault/var/www/openmediavault/js/omv/data/proxy/Rpc.js
@@ -69,13 +69,13 @@ Ext.define("OMV.data.proxy.Rpc", {
 		var me = this;
 		var request = me.buildRequest(operation);
 		Ext.apply(request, {
-			timeout: me.timeout,
 			scope: me,
 			callback: me.createRequestCallback(request, operation,
 			  callback, scope),
 			method: me.getMethod(request),
 			disableCaching: false
 		});
+		request.setTimeout(me.timeout);
 		OMV.Rpc.request(request);
 		return request;
 	},


deb/openmediavault/var/www/openmediavault/js/omv/grid/Panel.js
```diff
@@ -95,7 +95,10 @@ Ext.define("OMV.grid.Panel", {
 			var fnActivateTask = function() {
 				if (Ext.isEmpty(me.reloadTask)) {
 					me.reloadTask = Ext.util.TaskManager.newTask({
-						run: me.doReload,
+						run: function() {
+							if(!me.getStore().isLoading())
```
votdev: in my opinion this check should be done in doReload().

CyBHFal: It would be prettier but i wanted only to change autoreload feature behavior to avoid side effect with other calls to doReload (it is my first read of the source :-)). 

So I should do the checking in doReload() but in the derived class instead (omv/workspace/grid/Panel.js) ?

votdev: You can strip this commit from your PR because https://github.com/openmediavault/openmediavault/commit/7ae15b03ac177ee373055120c3314b10a747a7e2 will fix that issue.

votdev:
I think it is here: http://docs.sencha.com/extjs/6.2.0/classic/src/Base.js-7.html, e.g. in the start() method.

Please add a comment like 'Use setTimeout() here because Ext.data.request.Base uses the getter method to access the timeout config option.'

CyBHFal:
Line 24960 in ext-all-debug.js

`
    /**
     * Start the request.
     */
    start: function() {
        var me = this,
            timeout = me.getTimeout();
        if (timeout && me.async) {
            me.timer = Ext.defer(me.onTimeout, timeout, me);
        }
    },
`

Call stack:

start (ext-all-debug.js:24959)
callParent (ext-all-debug.js:12493)
start (ext-all-debug.js:25485)
request (ext-all-debug.js:26560)
callParent (ext-all-debug.js:12493)
**request (Rpc.js:82)
doRequest (Rpc.js:81)**

votdev:
Please sign your PR via ``Signed-off-by: Frank Mustermann <frank.mustermann@xxx.yyy>``. You also have to subscripe the CLA online.

deb/openmediavault/var/www/openmediavault/js/omv/grid/Panel.js
CyBHFal: Could you cherry-pick your correction https://github.com/openmediavault/openmediavault/commit/7ae15b03ac177ee373055120c3314b10a747a7e2 to 3.x in your next release ?
Thanks.

votdev: This is already done: https://github.com/openmediavault/openmediavault/commit/75a8a83755429b87c3e422dc58fa8282e229d420

votdev: Done, see https://github.com/openmediavault/openmediavault/commit/74273141a39e3be90498fd3fe3a66a40641da98e

PR #80, opened 2018-04-04
Make script of recyclebin delete working from cron

Fantu:
Scripts of recyclebin delete was working only executing them manually,
after many tests and changes seems now working also launched
automatically from cron.
- use bash instead dash (from system default on sh), same of user
setting where manually works
- added some lines to help debug (after can be removed)
- use file instead dir for lock inluding pid to check in it and stop
only if a previous script is really running
- remove lock always on end script and on all exit with trap


votdev:
What happens when the root user has dash as default shell? Then we have the same problem as now.

votdev:
Your solution to create the lock file is not atomic. We had some situations in the past where multiple scripts are executed at the same time which results in race conditions. Any comments on that?

Fantu:
lock with mkdir make me crazy with absurd result, for example where tried to do it failing because exists but are not and also with rm -rf that tell it not exists, lock file with pid seems solved all absurd cases and seems solved also the case where script "disappair" after start and without trap that remove the lock.
I logged all to files and for now seems all scripts start, end and do trap with exit 0 also auto from cron.
I not had time to check each thing of script to found and remove bashism (or with unexpected result on dash), and I not tried to change user default shell as on this system I have other things added not related to openmediavault things and I don't want risk on production system.
Since seems many users have the same issue I asked to the issue related help to test it, @votdev wait some days for more result from other users (I'll also keep checked the system that run it) to be sure is fully solved.

Sorry for my bad english.

enzolis:
votdev is correct. The file locking mechanism implemented is not atomic. The previous mkdir approach is probably the only atomic way if you want to do it with standard filesystem operations.
Another way would be using flock.
Cheers,
 Thimo

Samuel12321:
Suggest Close, last activity over a year ago, and issue it was for has been closed

PR #89, opened 2018-04-23
Added some missing French translations

vinceh121:
I saw that a bunch of French translations were missing, so I decided to translate some of them.


votdev:
I've added the translations manually on https://translate.openmediavault.org/.

Issue #113, opened 2018-05-27
Problem with Apply button - Failed to execute command


joakimlemb:
After mounting a new filesystem and trying to apply settings I get the following error:

```
Failed to execute command 'export PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin; export LANG=C; monit monitor collectd 2>&1' with exit code '1': Cannot create socket to [localhost]:2812 -- Connection refused
Error #0:
OMV\ExecException: Failed to execute command 'export PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin; export LANG=C; monit monitor collectd 2>&1' with exit code '1': Cannot create socket to [localhost]:2812 -- Connection refused in /usr/share/php/openmediavault/system/process.inc:175
Stack trace:
#0 /usr/share/php/openmediavault/system/monit.inc(115): OMV\System\Process->execute()
#1 /usr/share/php/openmediavault/system/monit.inc(82): OMV\System\Monit->action('monitor', false)
#2 /usr/share/openmediavault/engined/module/collectd.inc(81): OMV\System\Monit->monitor()
#3 /usr/share/openmediavault/engined/rpc/config.inc(194): OMVModuleCollectd->startService()
#4 [internal function]: OMVRpcServiceConfig->applyChanges(Array, Array)
#5 /usr/share/php/openmediavault/rpc/serviceabstract.inc(123): call_user_func_array(Array, Array)
#6 /usr/share/php/openmediavault/rpc/serviceabstract.inc(149): OMV\Rpc\ServiceAbstract->callMethod('applyChanges', Array, Array)
#7 /usr/share/php/openmediavault/rpc/serviceabstract.inc(565): OMV\Rpc\ServiceAbstract->OMV\Rpc\{closure}('/tmp/bgstatusgh...', '/tmp/bgoutputfh...')
#8 /usr/share/php/openmediavault/rpc/serviceabstract.inc(159): OMV\Rpc\ServiceAbstract->execBgProc(Object(Closure))
#9 /usr/share/openmediavault/engined/rpc/config.inc(213): OMV\Rpc\ServiceAbstract->callMethodBg('applyChanges', Array, Array)
#10 [internal function]: OMVRpcServiceConfig->applyChangesBg(Array, Array)
#11 /usr/share/php/openmediavault/rpc/serviceabstract.inc(123): call_user_func_array(Array, Array)
[...]

PR #115, opened 2018-05-30
Reduce startup delay from 30 to 0 for monit daemon

joakimlemb:
Recently discussed in https://github.com/openmediavault/openmediavault/issues/113
Should work around the timing issue with monit when openmediavault is trying to communicate with monit, but monit is not ready.


votdev:
I think 0 seconds is to optimistic. What about 5 or 10 seconds delay. Otherwise it may happen on various systems that the admin will get multiple emails from monit that some services are not up if they need some more time to setup after a reboot.

joakimlemb:
A good point, set what you feel comfortable with. (I have been running it at 0 for 2 days now without any noticeable issues.)

Another workaround for that would be setting the After parameter in the systemd file for monit to load last on boot.

votdev:
Fixed with https://github.com/openmediavault/openmediavault/commit/605ace9ad01f3d474a1e83058c6c83d6720b2930. Thx for the hints.

PR #129, opened 2018-07-07
Updates samba homes default configuration

npenin:
added execution permission so that tools can be stored in home folder


votdev:
I don’t think this is a good idea as a default setting. If you like to override this settings you should use the /etc/default/openmediavault file.

votdev:
Remember that the filesystem where the shared folders are located is mounted without execution flag.

npenin:
yes, I forgot to mention, that I wanted to have tools in my home folder, accessible through samba. to be executable on windows.

PR #162, opened 2018-08-26
Use systemd to configure the network.

votdev:


ryecoaaron:
Is systemd-resolved ever going to be more than just enabled/started at installation time?  If it is started in more places than just postinst, it will be very difficult to have a dnsmasq (or other dns server) plugin.

votdev:
It is only started in debian/postinst.

PR #166, opened 2018-08-29
grammar problem resolved.

emamirazavi:
grammar problem in description of the field resolved.


votdev:
Note, this changes can not be backported to 4.x becaus eotherwise we will loose all translations for this fields.

emamirazavi:
It's good idea to update keys of all translations.

PR #195, opened 2018-10-26
Use /bin/sh as default for new users

ThomasKaiser:
https://forum.openmediavault.org/index.php/Thread/24595-Fix-the-shell-Completion-history-and-prompt/?postID=186401#post186401

### DIFF
--- a/deb/openmediavault/var/www/openmediavault/js/omv/module/admin/privilege/user/User.js
+++ b/deb/openmediavault/var/www/openmediavault/js/omv/module/admin/privilege/user/User.js
@@ -119,7 +119,7 @@ Ext.define("OMV.module.admin.privilege.user.user.General", {
 				emptyText: _("Select a shell ..."),
 				valueField: "path",
 				displayField: "path",
-				value: "/bin/dash"
+				value: "/bin/sh"
 			},{
 				xtype: "checkbox",
 				name: "disallowusermod",


votdev:
OMV will still use Dash as default shell. See

* https://wiki.debian.org/Shell
* https://wiki.ubuntu.com/DashAsBinSh

If it is necessary to switch this behaviour this can be accomplished via WebUI.

If there are overwhelming or security related reasons to switch to something else than the Debian default we can discuss this.

votdev:
Switching to ``/bin/sh`` does not help because this is ALWAYS a symlink to Dash.

Ubuntu: (my shell is zsh here)
```
➜  ceph issue_36480 ✗ ls -alh /bin/sh
lrwxrwxrwx 1 root root 4 Jul 20 09:49 /bin/sh -> dash
```

Debian:
```
root@omv4box:/home/vagrant# ls -alh /bin/sh
lrwxrwxrwx 1 root root 4 Jan 24  2017 /bin/sh -> dash
root@omv4box:/home/vagrant# which bash
/bin/bash
```

ThomasKaiser:
As [written already on the forum](
https://forum.openmediavault.org/index.php/Thread/24595-Fix-the-shell-Completion-history-and-prompt/?postID=186401#post186401)...

* `/bin/sh` is the default shell (not `/bin/dash`)
* `/bin/sh` is by default a symlink to `/bin/dash` (which allows any admin to change this by linking it to something else and change behavior system wide)
* It is an UI/UX issue since `/bin/dash` will get confused with `/bin/bash` pretty easily

All I'm asking for is really doing what Debian does and that's using `/bin/sh` as default shell. This won't change anything wrt functionality since as we all know `/bin/sh` is usually just a symlink to `/bin/dash`

But it will allow users to spot it easily if they want to create users with `/bin/bash` instead since now they will most probably simply overlook that dash is not bash (this won't happen if the UI default is sh)

TL;DR: dash and bash look too similar and using `/bin/sh` will make OMV more compatible to what Debian does:

    tk@hermes:~$ lsb_release -c
    Codename:	stretch
    tk@hermes:~$ grep SHELL /etc/default/useradd
    # The SHELL variable specifies the default login shell on your
    SHELL=/bin/sh

PR #211, opened 2018-11-14
Change 99openmediavault-proxy mode to 0644

zcsizmadia:
/etc/apt/apt.conf.d/99openmediavault-proxy does not have read permission for non root users (0640, instead of 0644 like the other files), so some apt commands, which should not require root permission, fail (e.g. apt search)

### DIFF
--- a/deb/openmediavault/srv/salt/omv/deploy/apt/default.sls
+++ b/deb/openmediavault/srv/salt/omv/deploy/apt/default.sls
@@ -43,7 +43,7 @@ configure_apt_proxy:
     - template: jinja
     - user: root
     - group: root
-    - mode: 640
+    - mode: 644
 
 {% if not use_kernel_backports | to_bool %}
 


votdev:
I don't think this is a good idea because you will expose the proxy credentials incl. the password to the world.

If ``apt search`` fails then this is a design bug by the developers of the APT tools. I don't think that such a command needs to know the proxy credentials for a local search of the package database.

I am also asking myself why it is necessary to run this command as non privileged user on a NAS. Normal users shouldn't have access to the console on a NAS.

zcsizmadia:
OOps, that would explout the credentials for sure.
I regularly use my console for installing other non-OMV related packages, since the primary function of that machine is not being a NAS, so I use apt normally, and try not to run commands as sudo whenever as possible.

I think the issue with OMV is not using 0640 for the file in the conf.d folder, but creating 0 length file, even if the proxy is not set, causing apt (e.g. search) failing with non root access.

Is there a way no to create the 0 length files during deployment, only when they are configured?

zcsizmadia:
I never used salt before, so after reading about it, is the issue simply is surround the file definition in default.sls with condition , checking for proxy enabled?

votdev:
Fixed, see https://github.com/openmediavault/openmediavault/commit/1fcd711ba3c0be9963f64ce4ec5ca6455eefca76.

zcsizmadia:
Awesome support! Thanks!

PR #243, opened 2018-12-22
Display a warning box when creating a RAID

votdev:
Display a warning box when creating a RAID if devices are used that are connected via USB.

This solution is better in my opinion than the previous implementation which does not show USB devices in the creation dialog. Thus users can still create a RAID, but they are informed about the problem. Thus we do not lock them out (but it might happen that we still have to answer question about problems with RAID and USB devices).


ryecoaaron:
While it is probably the right thing to do, most people will still do it (since they are "testing") because it is easy.  My forum support side of me says we should force them to use the command line if they really want to make this bad decision.

PR #255, opened 2018-12-27
Add extra honor to the Dune novel

votdev:
Rename omv-salt to omv-spice because spice is the new salt and to add another extra honor to the Dune novel.

https://en.wikipedia.org/wiki/Melange_(fictional_drug)


votdev:
Too late, omv-salt jas been established in the meanwhile.

PR #323, opened 2019-03-16
Update mkconf/hosts to clarify hosts is generated by omv

emontnemery:
Update mkconf/hosts to clarify it's generated by omv

I spent a lot of time searching for who was overwriting the file, unfortunately I found this outdated forum post which led me in the wrong direction:
https://forum.openmediavault.org/index.php/Thread/1490-Setup-edit-etc-hosts/?postID=6822#post6822

Hopefully this can save someone else's time.

BTW:
The reason I need to edit `/etc/hosts` is because this line screws things up for dnsmasq:
`127.0.1.1 ${fqdn} ${alias}`


votdev:
Thx for the hint. I've used the header from OMV5.
Fixed with https://github.com/openmediavault/openmediavault/commit/3ddcc0527d47bd3821df43a661053db32d7bdbc7.

Issue #343, opened 2019-04-24
Standby feature conflicts with watchdog behavior


g-provost:
OMV installs watchdog service by default. When the system is put in standby / suspend mode watchdog is not stopped, therefore since nothing will be kicking the watchdog in that state, the watchdog will reset the suspended system after 60sec. 

This issue was reported by a user running OMV4 on Helios4... and we easily reproduced the issue. The work around given was to disable watchdog.service for now.

After digging and testing I'm surprise that the issue wasn't reported by other OMV users of other platform because the issue is not hardware platform related.

The issue applies whichever hardware watchdog or software watchdog (aka softdog) is used. BTW need to take note that watchdog app will use hardware watchdog if present even though _/etc/default/watchdog_  state to use softdog.

After insuring that the system uses softdog and not hw watchdog, I was expecting that in suspend mode the softdog will get also suspended since it is software... but it is not !? I'm still a bit puzzled here.

I'm still trying to figure out if systemctl could be expected to stop the watchdog.service before doing the suspend... and then restart the service when waking-up... or it needs to be handled by app (OMV) that triggers the suspend.

Also shouldn't OMV used systemd instead of pm-utilis since it's the recommend way for Debian distrib since Jessie ?

votdev:
> Also shouldn't OMV used systemd instead of pm-utilis since it's the recommend way for Debian distrib since Jessie ?

This is done in OMV5.

votdev:
If this can be fixed easily we can add a systemd unit file, but IMHO this is an upstream issue that should be finally fixed there.

g-provost:
> If this can be fixed easily we can add a systemd unit file, but IMHO this is an upstream issue that should be finally fixed there.

I don't think it can be considered as an upstream issue... because many people will argue that it's very unlikely to run a watchdog on a system that can go in suspend mode.

So my opinion is that it needs to be supported by the application (OMV) and with systemd is very simple. Just add the following **watchdog** script with executable right in folder **/lib/systemd/system-sleep**

```
#!/bin/sh

case $1 in
  pre)
    systemctl stop watchdog.service
  ;;
  post)
    systemctl start watchdog.service
  ;;
esac
```

For pm-utlis is actually as simple. Just add the following **99_watchdog** with executable right in folder **/usr/lib/pm-utils/sleep.d/**

```
#!/bin/sh

case "${1}" in
  suspend)
    systemctl stop watchdog.service
  ;;
  resume)
    systemctl start watchdog.service
[...]

PR #345, opened 2019-04-25
Stop watchdog.service when going suspend mode. Resume when waking up.

g-provost:
Fix issue #343

### DIFF
--- a/deb/openmediavault/lib/systemd/system-sleep/openmediavault
+++ b/deb/openmediavault/lib/systemd/system-sleep/openmediavault
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+# Watchog need to be stopped otherwise system will reset during suspend state
+case $1 in
+  pre)
+    if [ !$(systemctl is-active --quiet watchdog.service) ]; then
+      systemctl stop watchdog.service
+    fi
+  ;;
+  post)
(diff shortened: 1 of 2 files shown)


votdev:
LGTM, but i need to check why pm-utils is still a dependency. Maybe we can get rid of it.

deb/openmediavault/lib/systemd/system-sleep/watchog
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
votdev: Can you please rename this file to ``openmediavault-watchdog``, so it is clear that this file comes from OMV. Same applies to the pm-utils file.

g-provost: Ok, then let's call each script just `openmediavault` since maybe in the feature you might add more tweak that are not necessary related to watchdog. OK ?

Also I will add states hibernate+thaw to the script for pm-utilis since the same use case apply for hibernate state.

votdev:
Please rebase and remove the pm-utils code.
pm-utils is not used anymore in OMV, see https://github.com/openmediavault/openmediavault/pull/369.

votdev:
Superseded by https://github.com/openmediavault/openmediavault/pull/372.

PR #377, opened 2019-05-15
Prefer X25519 over NIST P-384 and NIST P-256

jannispinter:
Prefer X25519 over NIST P-384 and NIST P-256 in nginx's configuration for the webgui. Compared to my previous PR, this one also includes P-256 as a fallback option for older TLS clients.


votdev:
Superseded by https://github.com/openmediavault/openmediavault/commit/de71e70519ca6255934e77a799cc22f025517621 because the PR does not match the requirements, e.g. the signed-of-by is missing. Anyway thx for the contribution.

PR #381, opened 2019-05-26
Run clamdscan as root

votdev:
Run clamdscan as root to give access to all shared folders without user interaction.

Without this it is necessary that the administrator gives the user or group ``clamav`` the correct ACL permissions (at least ``read``) recursively to the shared folder to be processed by ClamAV, otherwise the daemon process can not access the directories/files to process.

Running the process as root will reduce the complexity of configuring the correct privileges for the admin. Applying the privileges automatically is not wanted because this is too complex, will increase failures and modifies the privileges set by the administrator without his knowledge. On the other side running a daemon as root is a security issue and should be prevented when possible.

Example of the problems raised by this issue:
https://forum.openmediavault.org/index.php/Thread/27298-clamav-error-Could-not-connect-to-clamd-on-LocalSocket-var-run-clamav-clamd-ctl/

- [ ] References issue


votdev:
Can not reproduce this anymore. It seems ClamAV requires some time to settle up after the scan job has been created.

PR #404, opened 2019-07-04
update omv-salt for deploying failed

yyldavidj:
The second param should be 'list' for function salt.client.Caller.cmd, otherwise it will prompt
'No matching sls found for '<map object at xxx>' in env 'base''
Test with 'salt 2019.2.0'

- [ ] References issue

### DIFF
--- a/deb/openmediavault/usr/sbin/omv-salt
+++ b/deb/openmediavault/usr/sbin/omv-salt
@@ -117,7 +117,7 @@ def deploy_run(names, quiet):
     names = map(lambda x: "omv.deploy.{}".format(x), names)
     mopts = salt.config.minion_config(__salt_minion_config_file)
     caller = salt.client.Caller(mopts=mopts)
-    result = caller.cmd("state.orchestrate", names)
+    result = caller.cmd("state.orchestrate", list(names))
     rc = result["retcode"]
     # Print the output.
     if not quiet:


votdev:
I'm wondering why this is necessary because ``names`` is already a ``list``. By the way, OMV uses Salt 2018.3.4, i'm wondering how you get 2019.2.0 running on Debian 10 because Saltstack only provides packages for Debian 9 which has install dependency issues.

yyldavidj:
@votdev 

> I'm wondering why this is necessary because `names` is already a `list`. 

`names` is an `iterator ` rather than `list` as return value of function `map` in python 3.
https://docs.python.org/3/library/functions.html?highlight=map#map

Please pay attention to the differences between python 3 & 2.
The type of Function `map`'s return value is `list` in python 2.
https://docs.python.org/2.7/library/functions.html?highlight=map#map

> By the way, OMV uses Salt 2018.3.4, i'm wondering how you get 2019.2.0 running on Debian 10 because Saltstack only provides packages for Debian 9 which has install dependency issues.

I'm using Ubuntu 18.04.02 LTS, and I got 2019.2.0 via pypi.org
i.e. `pip3 install salt`
Indeed, there are some dependency issues, but they can be solved by installing extra packages, such as
`apt install gcc g++ python3-dev`

PR #432, opened 2019-09-05
Modify shared folder unit file

votdev:
Change the targets to ensure all filesystems are ready and up.

https://forum.openmediavault.org/index.php/Thread/28550-Shared-Folders/
https://forum.openmediavault.org/index.php/Thread/25042-MergerFS-folders-not-mounted-in-sharedfolders/

- [ ] References issue

### DIFF
--- a/deb/openmediavault/srv/salt/omv/deploy/systemd/10sharedfolders.sls
+++ b/deb/openmediavault/srv/salt/omv/deploy/systemd/10sharedfolders.sls
@@ -17,6 +17,9 @@
 # You should have received a copy of the GNU General Public License
 # along with OpenMediaVault. If not, see <http://www.gnu.org/licenses/>.
 
+# Documentation/Howto:
+# https://www.freedesktop.org/software/systemd/man/bootup.html#System%20Manager%20Bootup
+
 {% set sharedfolders_path = salt['pillar.get']('default:OMV_SHAREDFOLDERS_DIR', '/sharedfolders') %}
 {% set sharedfolders_path_escaped = salt['cmd.run']('systemd-escape --path ' ~ sharedfolders_path) %}
 {% set sharedfolders = salt['omv_conf.get']('conf.system.sharedfolder') %}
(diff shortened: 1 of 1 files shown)


ryecoaaron:
This will break sharedfolders on zfs.

votdev:
> This will break sharedfolders on zfs.

I don't think so because ZFS things are done before ``local-fs.target``, see https://github.com/zfsonlinux/zfs/blob/master/etc/systemd/system/zfs-mount.service.in#L8. At ``sysinit.target`` everyting related to mounting should be done (according to https://www.freedesktop.org/software/systemd/man/bootup.html#System%20Manager%20Bootup).

ryecoaaron:
> I don't think so because ZFS things are done before local-fs.target

They are supposed to be done before local-fs.target but they aren't.  Hopefully they are by the sysinit time frame.

I'm still not sure how this fixes the mount during apply though.  I understand it would fix the mount during boot issue.  I guess I will do some testing.

I know this is a different issue but this doesn't fix the remount triggered during sharedfolder creation that fails to restart a mount unit because the folder is in use.

votdev:
> > I don't think so because ZFS things are done before local-fs.target
> 
> They are supposed to be done before local-fs.target but they aren't. Hopefully they are by the sysinit time frame.
> 

If not, then we can simply re-add the ``After=zfs-mount.service``line.

> I'm still not sure how this fixes the mount during apply though. I understand it would fix the mount during boot issue. I guess I will do some testing.
> 

It won't, but it will fix the booting issue. A small step forward.

subzero79:
Another approach in my opinion could be use a custom target line up with graphical or multi-user. We know multi-user is the last one maybe omv could set the default sharedfolder.target doesn’t do any harm. 
Cons:could hurt cron scripts that are set with the @reboot instead of cron time

votdev:
Should be finally fixed by f5e4d283eff11ab291e7fa0b5a4fd0aea2735d9d. It seems the order of the conditions was the key.

PR #433, opened 2019-09-05
Modify shared folder unit file

votdev:
Change the targets to ensure all filesystems are ready and up.

https://forum.openmediavault.org/index.php/Thread/28550-Shared-Folders/
https://forum.openmediavault.org/index.php/Thread/25042-MergerFS-folders-not-mounted-in-sharedfolders/

- [ ] References issue


geoffae:
I can confirm this works 👍

Issue #444, opened 2019-09-10
Additional servers on nginx using the same ports do not work cause of the use of ipv6only and deferred parameters


EzequielB:
### Additional servers on nginx using the same ports do not work cause of the use of ipv6only and deferred parameters
Since `ipv6only` and `deferred` parameters forces a new bind call on nginx http://nginx.org/en/docs/http/ngx_http_core_module.html, other servers can't be set on the same ports.
The use case here is having OMV as the main linux distro and adding custom servicies on top that would reuse nginx service.
Removing the ipv6only and adding both ipv4 and ipv4 listen configurations remediates the use of ipv6only flag.

### Steps to reproduce issue
Add a new server on the nginx configuration inside site-enabled (for example a default server)
´´´
server {
	listen 80 default_server;
	listen [::]:80 default_server;

        if ($scheme = http) {
            # Force redirection to HTTPS.
            return 301 https://$host:443$request_uri;
        }
        
        listen 443 ssl default_server;
        listen [::]:443 ssl default_server;
        
        ssl_certificate [SOME_CERT];
        ssl_certificate_key [SOME_KEY];
}
´´´
Try to restart nginx.

### Versions report
4.x

votdev:
> The use case here is having OMV as the main linux distro and adding custom servicies on top that would reuse nginx service.

Please use Docker for that.

The current implementation of the nginx config will not be changed anymore because there are reasons we do it this way. Check the history of the file and you'll see that there are many changes related this thing because it causes errors on other systems if we do it NOT that way.

EzequielB:
The idea is to use docker and have nginx as a forward proxy. With the current state, you just can't listen on the same ports (80 or 443) with different domains or paths.
Maybe in v5 it can be modified, let me know if I can help there.
Thanks for the reply.

votdev:
As said, i don't want to change such issues in the stable branch. OMV5 is in beta phase, so this is ok to do such changes that might introduce regressions.

votdev:
> Maybe in v5 it can be modified, let me know if I can help there.

This would be great if you could provide a PR for OMV5.

ryecoaaron:
> With the current state, you just can't listen on the same ports (80 or 443) with different domains or paths.

[...]

PR #445, opened 2019-09-10
Remove parameters forcing a new bind on nginx

EzequielB:
Remove parameters forcing a new bind on nginx

When a new server is added manually in nginx configuration, the same ports can't be used cause of the ipv6only and deferred parameters used.
http://nginx.org/en/docs/http/ngx_http_core_module.html

Fixes: https://github.com/openmediavault/openmediavault/issues/444

- [ ] References issue


votdev:
See my comments on #444. The nginx configuration is tricky and has caused many troubles (see changelog and history of the omv-mkconf script). We can give it a try in OMV5, but OMV4 is closed for such things (never touch a running system).

Issue #446, opened 2019-09-11
Syslog is flooded with messages complaining about smtpd_relay_restrictions and smtpd_recipient_restrictions


jimmymasaru:
### Description of issue/question
Syslog is flooded with messages complaining about smtpd_relay_restrictions and smtpd_recipient_restrictions

### Steps to reproduce issue
- Set up notification
- Wait for a while. A system restart might help speeding up.
- Check syslog
- Syslog is flooded with the following message
> fatal: in parameter smtpd_relay_restrictions or smtpd_recipient_restrictions, specify at least one working instance of: reject_unauth_destination, defer_unauth_destination, reject, defer, defer_if_permit or check_relay_domains

### Versions report
5.0 beta

votdev:
A better solution is to set compatibility_level to 2, thus the default config of Postfix will be used. See https://github.com/openmediavault/openmediavault/commit/2ab78f559da88e1683b8cc7d2c086d3dfaaba0cf.

PR #447, opened 2019-09-11
Add smtp_relay_restrictions with explicit default value to the postfix config

jimmymasaru:
Add smtp_relay_restrictions with explicit default value to the postfix config

If the value of smtp_relay_restrictions is empty, postfix keeps flooding syslog with fatal error messages. We should explicitly add the default value of smtp_relay_restrictions to the configuration file to suppress the error messages. 
[Reference](http://www.postfix.org/postconf.5.html#smtpd_relay_restrictions)

Fixes: #446

Signed-off-by: Jimmy Page <jimmymasaru@gmail.com>


votdev:
Better solution is to force Postfix to use it's default value. This can be done by setting compatibility level to 2. Done with https://github.com/openmediavault/openmediavault/commit/2ab78f559da88e1683b8cc7d2c086d3dfaaba0cf.

```
# postconf | grep compatibility_level
append_dot_mydomain = ${{$compatibility_level} < {1} ? {yes} : {no}}
compatibility_level = 0
mynetworks_style = ${{$compatibility_level} < {2} ? {subnet} : {host}}
relay_domains = ${{$compatibility_level} < {2} ? {$mydestination} : {}}
smtpd_relay_restrictions = ${{$compatibility_level} < {1} ? {} : {permit_mynetworks, permit_sasl_authenticated, defer_unauth_destination}}
smtputf8_enable = ${{$compatibility_level} < {1} ? {no} : {yes}}
```

PR #469, opened 2019-10-17
fix: flush_interval was not replaced

long-walk:
flush_interval was not replaced in /etc/default/rrdcached
I hop tihs is ok with no CLA


long-walk:
sorry still wrong syntax

PR #470, opened 2019-10-20
correct syntax for flush_interval in rrdcached/default.sls

long-walk:
flush_interval was not replaced in /etc/default/rrdcached correct syntax
I hop tihs is ok with no CLA


votdev:
A PR can’t be merged without a signed CLA.

long-walk:
ah, sorry I dint see that I can do it with two clicks. I read a something about print letters and mail it on this page: https://www.openmediavault.org/contributoragreement.html

Issue #82, opened 2018-04-07
Inappropriate IOCTL


EruIluvatar:
I just got a mail with the content:
> /etc/cron.weekly/openmediavault-update-smart-drivedb:
> mesg: ttyname failed: Unpassender IOCTL (I/O-Control) für das Gerät

The forum suggests a fix for `/root/.profile`.
But could the problem also be avoided by not executing the script as a login shell, by making the shebang `#! /bin/sh` instead of `#! /bin/sh -l` like here?

https://github.com/openmediavault/openmediavault/commit/3b080600ff2d23e12c0068d8a51be785649e1a87#diff-3d0f72f14b9a4eb86127299f58aec433

votdev:
Please check https://forum.openmediavault.org/index.php/Thread/20966-mesg-ttyname-failed-Inappropriate-ioctl-for-device/?postID=163028#post163028 how to fix that.

votdev:
Fixed by https://github.com/openmediavault/openmediavault/commit/bec9ba52cbde3fc3ef32435aa66ca2e14dbc3c1b in OMV5.

PR #471, opened 2019-10-21
Prevent TTY-related pseudo-errors

xBelladonna:
Prevents the TTY error "mesg: ttyname failed: Inappropriate ioctl for device" by not allocating login shell.

These message are sent out as mail as part of the cron job to update the smartdb, which the user gets weekly if mail is enabled. Reproducing this bug is as easy as enabling mail and calling the cron file /etc/cron.weekly/openmediavault-update-smart-drivedb in the `Scheduled Jobs` web interface tab if you don't want to wait a week.
We do this by configuring the Vagrant box to not allocate a pseudo-tty and issue inline commands, which leads to shell scripts not asking for login shell.

Fixes: #82 

Signed-off-by: xBelladonna <isabelladonnamoore@outlook.com>

- [x] References issue

### DIFF
--- a/vagrant/Vagrantfile
+++ b/vagrant/Vagrantfile
@@ -23,6 +23,9 @@
 hostname = "omv5box"
 
 Vagrant.configure("2") do |config|
+  # Prevent TTY-related pseudo-errors
+  # We do this by not allocating a pseudo-tty (the default is bash -l)
+  config.ssh.shell = "bash -c 'BASH_ENV=/etc/profile exec bash'"
 	config.vm.box = "debian/buster64"
 	config.vm.define hostname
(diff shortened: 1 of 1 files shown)


votdev:
I think https://github.com/openmediavault/openmediavault/commit/bec9ba52cbde3fc3ef32435aa66ca2e14dbc3c1b is the better approach to fix that because it does not only apply to the Vagrant setup.

votdev:
Fixed by https://github.com/openmediavault/openmediavault/commit/bec9ba52cbde3fc3ef32435aa66ca2e14dbc3c1b in OMV5.

xBelladonna:
Oh yes that looks like a better way to do it. Thank you!

PR #474, opened 2019-10-24
fix: diskstat do not install disk.sls file

long-walk:
fix: diskstat do not install disk.sls
not sure if something else is needed but on my test it generates the picture


votdev:
Thx

Issue #486, opened 2019-11-01
Use gender-neutral language when referring to users.


Julian-O:
### Description of issue/question

When referring to the user, the pronoun "his" is used. It is considered more inclusive to use gender-neutral pronouns when referring to users.

(Pull request to is following shortly.)

### Steps to reproduce issue

* In the WebGUI, select Access Rights Management | Users | Add

* Note the text "Disallow the user to modify **his** account.

Note: It would be an improvement to reword this to "**Allow** the user to..." and reverse the default to avoid double-negatives, but that has impact on translations.

### Versions report

4.1.26-1

votdev:
I thought ``his`` is gender neutral in english.

ryecoaaron:
Nope.  Even using his/hers is not considered all inclusive.  It should be "the user update their ...".

VoxelTek:
Unfortunately, this still seems to be present in the latest OpenMediaVault versions, as can be seen here on an instance of version 6.5.0-1. Could this be reopened, perhaps? @votdev 
![image](https://github.com/openmediavault/openmediavault/assets/53562267/b81c9ba8-6369-42e3-a309-fdd81cd65177)

ryecoaaron:
I just did a grep of all of the source code and this seems to be the only place left.

PR #487, opened 2019-11-01
Use gender-neutral language

Julian-O:
Particularly in the UI.

English text (only) changed, but it was a key into the translation files, so they were correspondingly updated.

My IDE automatically removes trailing blanks on lines of code, so a couple of those were corrected too.

(Confession: I don't have unit-tests configured locally. Relying on automated run.)

Fixes: #486

Signed-off-by: Julian-O <openmediavaultcorrespondence@somethinkodd.com>

- [X] References issue

### DIFF
--- a/deb/openmediavault/usr/share/openmediavault/engined/rpc/usermgmt.inc
+++ b/deb/openmediavault/usr/share/openmediavault/engined/rpc/usermgmt.inc
@@ -451,7 +451,7 @@ finish:
 	 *   \em password The plain password to use.
 	 *   \em email The users email address.
 	 *   \em comment Any text string. This field is optional.
-	 *   \em disallowusermod Set to FALSE to disallow the user to modify his
+	 *   \em disallowusermod Set to FALSE to disallow the user to modify their
 	 *     account.
 	 *   \em sshpubkeys The users SSH public keys.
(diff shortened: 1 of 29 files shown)


votdev:
@Julian-O There is no need to modify the .po files, your changes will get lost when i rebuild the languages. You only need to modify the text in the .(inc|js|php|...) files. The .pot will be generated out of this sources, then uploaded to Transifex (https://translate.openmediavault.org), the translators update their language there, and a script will download this .po files into the project again.

votdev:
Thx

Julian-O:
Re: .po files

Ah, I understand now. Thanks. I was trying to avoid making extra work for each of the translators - I was worried they would need to restranslate the phrase when their old translation is probably perfectly fine, just because the key changed.

votdev:
> Re: .po files
> 
> Ah, I understand now. Thanks. I was trying to avoid making extra work for each of the translators - I was worried they would need to restranslate the phrase when their old translation is probably perfectly fine, just because the key changed.

They need to do so, but the Transifex UI is intelligent enough to suggest the previous translation.

ryecoaaron:
Quite a few non-english languages will not need to be translated since the current form is considered gender neutral.

PR #541, opened 2019-12-23
Raspberry Pi WatchDog Timer Fix

RFBomb:
Raspberry Pi Watchdog Fix

Apologies for formatting or if I'm doing this incorrect, this is my first 'pull' request. 

OpenMediaVault uses the 'softdog' watchdog module by default. Raspian does not support this, as 'softdog' is a missing module. The reason the Raspian group will not provide it is likely due to the fact that there is a hardware watchdog built into the board that can be utilized. The proposed changes below allow use of that hardware watchdog to enable it to perform a full system reboot on system hang (whereas before the watchdog daemon would be errored out, and a system hang would result in someone walking over and pulling the power cable out of the Pi for a hard reset.)

Forum Link for my fix: https://forum.openmediavault.org/index.php/Thread/30063-Raspberry-Pi-4-Watchdog-Not-Working/?postID=223220#post223220


RFBomb:
Accidentally made it. IDK what I'm doing.

Issue #569, opened 2020-01-18
NFS: Kerberos (GSS) authenticated exports cannot be mounted by clients


Aearsis:
## Description of issue/question

Due to implicit `sec=sys` option on the NFSv4 root, kerberized exports do not have the root defined, and mounting such exports fails.

### Steps to reproduce issue

Define NFS export with `sec=krb5`, configure kerberos properly. Even then mounting fails:

```
# mount nas:/Share /mnt/nas -o sec=krb5,vers=4.2 -vv
mount.nfs4: timeout set for Sat Jan 18 22:36:18 2020
mount.nfs4: trying text-based options 'sec=krb5,vers=4.2,addr=1.2.3.4,clientaddr=5.6.7.8'
mount.nfs4: mount(2): Operation not permitted
mount.nfs4: Operation not permitted
```

This can be fixed simply by adding the `sec=krb5` to the root export options.

votdev:
You can customize that by adding `OMV_NFSD_V4_DEFAULT_EXPORT_OPTIONS="ro,fsid=0,root_squash,no_subtree_check,hide,sec=krb5p:krb5i:krb5:sys"` to `/etc/default/openmediavault` and run
```
# omv-salt stage run prepare
# omv-salt deploy run nfs
```

Aearsis:
Sure, I even tried it, but did not run the `stage run prepare` and it did not work. Also, I had to read the source code to know which variable to change, which is rather suboptimal :)

Couldn't the option be accessible from the web gui? There is plenty of space on the NFS settings tab.

votdev:
> Couldn't the option be accessible from the web gui? There is plenty of space on the NFS settings tab.

No, OMV will not become something like Webmin where you can configure everything.

Aearsis:
It just seems weird to me that you can directly specify options for exports, but not for the root.

In the beginning of setting up, I have added the option there manually, and made everything work. After a reboot, it took me quite long to realize what was overwritten to make it work again. I'm just trying to prevent others from making the same mistakes - if you think that current state is OK, just close this issue and PR, no problem on my side.

mi-hol:
closing due to votdev's   "no"  response

PunisherHD92:
In case anyone like me is googling this issue for OMV8 and finding this page, it looks like the option name has changed in /etc/default/openmediavault.
The new line is:
[...]

PR #570, opened 2020-01-18
Allow users to define kerberos-secured NFS exports

Aearsis:
Without proper `sec=krb5` the user cannot define kerberized exports, as
then the NFSv4 root seemingly does not exist. By adding the universal
`sec=krb5p:krb5i:krb5:sys` option to defaults we allow both kerberized and
host-authenticated exports to be created.

Fixes: #569 

Signed-off-by: Ondřej Hlavatý <aearsis@eideo.cz>


votdev:
Thx for the contribution, but i'm not sure if this options will break other installations if this will become the default setting.

votdev:
I will close this PR because it seems to have the potential to break other installations. If you want to use Kerberos you should also change the environment variable `OMV_NFSD_NEEDSVCGSSD=yes`. OMV provides the ability to customize everything to allow you to adapt it to your needs, thus this PR is not needed, too.

PR #581, opened 2020-01-30
add ubuntu support

ryecoaaron:
add ubuntu support

I have been testing OMV 5.x on Ubuntu 19.10 Eoan and it works very well.  The only issue I have found is that the backports and security repos are hardcoded to Debian repos.

Fixes: ubuntu support

Signed-off-by: Aaron Murray <plugins@omv-extras.org>

- [ ] References issue

### DIFF
--- a/deb/openmediavault/srv/salt/omv/deploy/apt/default.sls
+++ b/deb/openmediavault/srv/salt/omv/deploy/apt/default.sls
@@ -94,19 +94,23 @@ remove_apt_sources_list_debian_security:
   file.absent:
     - name: "/etc/apt/sources.list.d/openmediavault-debian-security.list"
 
-# Check if the Debian security repository is already configured (e.g.
+remove_apt_sources_list_os_security:
+  file.absent:
+    - name: "/etc/apt/sources.list.d/openmediavault-os-security.list"
+
+# Check if the Debian or Ubuntu security repository is already configured (e.g.
 # in /etc/apt/sources.list). Only add it if this is not the case.
 {% set repos = [] %}
 {% for value in salt['pkg.list_repos']().values() %}
 {% set _ = repos.extend(value) %}
 {% endfor %}
-{% if repos | rejectattr('disabled') | selectattr('type', 'equalto', 'deb') | selectattr('uri', 'match', '^https?://security.debian.org/debian-security$') | list | length == 0 %}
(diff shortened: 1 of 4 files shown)


myxal:
Apologies for abusing this issue to ask a question, but - @ryecoaaron, was that "OMV on Ubuntu" setup you had just built from source, or installed via package manager? I see the docs still say debian is required, and having tried to install OMV the way it's installed on existing debian installation - add key, repository, and install a package - I've hit the age-old issue of OMV package depending on a virtual `phpapi` package from debian that ubuntu doesn't have.

ryecoaaron:
It was just some testing. Ubuntu is not supported.

mi-hol:
> I've hit the age-old issue of OMV package depending on a virtual `phpapi` package from debian that ubuntu doesn't have.

Php-imagick seems to have had the same issue.
    As I understand it their solution was to use different virtual phpapi packages for Debian (phpapi-20180731) and Ubuntu (phpapi-20190902). I got this understanding from  [RepoScope(Linux repositories inspector)](https://reposcope.com/package/php-imagick/dependencies)
Did I get this right?

ryecoaaron:
It isn't hard to compile packages for a different php api but OMV would have to add different repos for different api versions.

votdev:
OMV will stay on Debian, there will be no official support for Ubuntu because there are too many differences that are hard to maintain. Users that want to run OMV on Ubuntu need to compile the required packages themself.

mi-hol:
> no official support for Ubuntu because there are too many differences that are hard to maintain

I understand that for past versions of Ubuntu, but for Ubuntu 21 Canonical seems to be more focused on compatibility with Debian than in the past. 
To explore if OMV could be running on Ubuntu 21, would a switch of the currently obvious blocker package "php-pam" to a PECL supported version be acceptable?

ryecoaaron:
Even if Ubuntu was going to be supported, a non-LTS release should never be targeted since they are only supported for 9 months where LTS releases are supported for 5+ years.

mi-hol:
> a non-LTS release should never be targeted

sure, but to test internally another release might be required as it seems Canonical changed directions for compatibility only recently (after receiving a lot of criticism for breaking it in the past)

ryecoaaron:
Why would you test a different version?  That isn't testing if it isn't targeting the desired version.  And who would be testing??

Unless you are referring to snap or the desktop, I don't know what this direction change or compatibility breaking stuff is about.

votdev:
For the moment OMV will stay on Debian only, because there is no benefit of supporting Ubuntu. OMV was never planned to support multiple platforms, it was tied to Debian from beginning.
Ubuntu LTS support does not bring any benefit to OMV because OMV has no LTS support, each version is based on the latest Debian version.

PR #584, opened 2020-02-01
RTL support for OMV UI

myomv100:
Hello 

This quick UI addition for the languages that are RTL.


myomv100:
I think the implementation could be better. This might be taken as a reference to what need to be done to support RTL. I might  redo it in a cleaner way soon.

vagrant/Vagrantfile
```diff
@@ -47,7 +47,7 @@ Vagrant.configure("2") do |config|
 	config.vm.provider :libvirt do |prov|
 		prov.memory = "2048"
 		# Add some disks.
-		for i in 0..2 do
+		for i in 0..1 do
```
votdev: See above

deb/openmediavault/var/www/openmediavault/css/theme-triton-rtl.scss
```diff
@@ -0,0 +1,292 @@
+/**
```
votdev: Please extract only the necessary changes here. This way there is no duplicate code and the changes that are needed for RTL are visible.

deb/openmediavault/var/www/openmediavault/css/theme-all-rtl.scss
```diff
@@ -0,0 +1,624 @@
+/**
```
votdev: Please extract CSS that is necessary for RTL. Prevent duplicate code.

Issue #583, opened 2020-02-01
Invalid RPC response.


Dude234:
### Description of issue/question
the inclusion of the data pool (unionfilesystem) is not possible

### Steps to reproduce issue
After setting up the Snapraid network I set up the unionfilesystem "data pool".
When I select the desired target under CIFS and choose the data pool for sharing, I get the mentioned error.

```
Error #0: 
OMV\Rpc\Exception: Invalid RPC response. Please check the syslog for more information. in /usr/share/php/openmediavault/rpc/rpc.inc:187
Stack trace:
#0 /usr/share/php/openmediavault/rpc/proxy/json.inc(97): OMV\Rpc\Rpc::call('ShareMgmt', 'getCandidates', Array, Array, 3)
#1 /var/www/openmediavault/rpc.php(45): OMV\Rpc\Proxy\Json->handle()
#2 {main}
```

```
Feb  1 15:47:36 openmediavault5 omv-engined[1753]: PHP Fatal error: Uncaught Error: Call to a member function getImpl() on null in /usr/share/php/openmediavault/system/filesystem/filesystem.inc:898
Feb  1 15:47:36 openmediavault5 omv-engined[1753]: Stack trace:
Feb  1 15:47:36 openmediavault5 omv-engined[1753]: #0 /usr/share/openmediavault/engined/rpc/sharemgmt.inc(88): OMV\System\Filesystem\Filesystem::getImplByMountPoint('/srv/5accee2a-9...')
Feb  1 15:47:36 openmediavault5 omv-engined[1753]: #1 [internal function]: Engined\Rpc\ShareMgmt->getCandidates(Array, Array)
Feb  1 15:47:36 openmediavault5 omv-engined[1753]: #2 /usr/share/php/openmediavault/rpc/serviceabstract.inc(123): call_user_func_array(Array, Array)
Feb  1 15:47:36 openmediavault5 omv-engined[1753]: #3 /usr/share/php/openmediavault/rpc/rpc.inc(86): OMV\Rpc\ServiceAbstract->callMethod('getCandidates', Array, Array)
Feb  1 15:47:36 openmediavault5 omv-engined[1753]: #4 /usr/sbin/omv-engined(537): OMV\Rpc\Rpc::call('ShareMgmt', 'getCandidates', Array, Array, 1)
Feb  1 15:47:36 openmediavault5 omv-engined[1753]: #5 {main}
[...]

Issue #582, opened 2020-01-31
Invalid RPC Response while adding a new shared folder


coffinbearer:
Similar to #550 i get an Invalid RPC Responce error in the device selection during adding a shared folder from a btrfs 4-drive pool while opening the select device dropdown.
I was able to add shared folders from a single btrfs drive.
fresh install of openmediavault:
Release: 5.2.6-1
Linux 5.4.0-0.bpo.2-amd64 #1 SMP Debian 5.4.8-1~bpo10+1 (2020-01-07) x86_64 GNU/Linux

Syslog:
`Jan 31 18:28:45 omv28756 omv-engined[2974]: PHP Fatal error: Uncaught Error: Call to a member function getImpl() on null in /usr/share/php/openmediavault/system/filesystem/filesystem.inc:898
Jan 31 18:28:45 omv28756 omv-engined[2974]: Stack trace:
Jan 31 18:28:45 omv28756 omv-engined[2974]: #0 /usr/share/openmediavault/engined/rpc/sharemgmt.inc(88): OMV\System\Filesystem\Filesystem::getImplByMountPoint('/srv/dev-disk-b...')
Jan 31 18:28:45 omv28756 omv-engined[2974]: #1 [internal function]: Engined\Rpc\ShareMgmt->getCandidates(Array, Array)
Jan 31 18:28:45 omv28756 omv-engined[2974]: #2 /usr/share/php/openmediavault/rpc/serviceabstract.inc(123): call_user_func_array(Array, Array)
Jan 31 18:28:45 omv28756 omv-engined[2974]: #3 /usr/share/php/openmediavault/rpc/rpc.inc(86): OMV\Rpc\ServiceAbstract->callMethod('getCandidates', Array, Array)
Jan 31 18:28:45 omv28756 omv-engined[2974]: #4 /usr/sbin/omv-engined(537): OMV\Rpc\Rpc::call('ShareMgmt', 'getCandidates', Array, Array, 1)
Jan 31 18:28:45 omv28756 omv-engined[2974]: #5 {main}
Jan 31 18:28:45 omv28756 omv-engined[2974]:   thrown in /usr/share/php/openmediavault/system/filesystem/filesystem.inc on line 898`

WebUI:
`Error #0:
OMV\Rpc\Exception: Invalid RPC response. Please check the syslog for more information. in /usr/share/php/openmediavault/rpc/rpc.inc:187
Stack trace:
[...]

PR #585, opened 2020-02-02
Fix invalid RPC response in file sharing for non local mounts

Helly1206:
Pull request for issues:
[https://github.com/openmediavault/openmediavault/issues/583](https://github.com/openmediavault/openmediavault/issues/583)
and
[https://github.com/openmediavault/openmediavault/issues/582](https://github.com/openmediavault/openmediavault/issues/582)

- [x] References issue

### DIFF
--- a/deb/openmediavault/usr/share/php/openmediavault/system/filesystem/filesystem.inc
+++ b/deb/openmediavault/usr/share/php/openmediavault/system/filesystem/filesystem.inc
@@ -895,6 +895,20 @@ class Filesystem extends \OMV\System\BlockDevice
 		$deviceFile = $mp->getDeviceFile();
 		$mngr = Backend\Manager::getInstance();
 		$backend = $mngr->getBackendById($deviceFile);
+		// if no backend found, try finding type and get backend by type
+		if (NULL == $backend) {
+			$type = "";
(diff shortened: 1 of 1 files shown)


votdev:
The final solution to fix that issue is to find out why the existing implementation does not work. This PR is only a workaround. The problem is surely BTRFS which handles many things different than other filesytems. That's the first thing that should be investigated.

deb/openmediavault/usr/share/php/openmediavault/system/filesystem/filesystem.inc
```diff
@@ -895,6 +895,20 @@ class Filesystem extends \OMV\System\BlockDevice
 		$deviceFile = $mp->getDeviceFile();
 		$mngr = Backend\Manager::getInstance();
 		$backend = $mngr->getBackendById($deviceFile);
+		// if no backend found, try finding type and get backend by type
+		if (NULL == $backend) {
+			$type = "";
+			foreach ($mngr->enumerate() as $datak => $datav) {
```
votdev: The enumerate function is too expensive in this case. On slow systems it can take ages to collect the necessary data.

votdev:
Closed in favor of https://github.com/openmediavault/openmediavault/pull/587.

PR #594, opened 2020-02-08
adding tags-like styling for user groups in User.js

abo-khalid:
Hello

This is a proposed appearance modification to make user groups visually easy to read.
 
Screenshot from 2020-02-08 08-44-36

### DIFF
--- a/deb/openmediavault/var/www/openmediavault/css/theme-all.scss
+++ b/deb/openmediavault/var/www/openmediavault/css/theme-all.scss
@@ -621,3 +621,14 @@ div#headerrlogo {
 .x-webkit :not(.x-form-textarea-body) > .x-form-trigger-wrap {
 	height: initial;
 }
+
+/******************************************************************************/
+/* Tags class for group cell in User grid panel */
+.x-group-tag {
+	display: inline-block;
+	border: 1px solid #d0d0d0;
+	padding: 0px 3px 0px 3px;
+	margin: 1px;
+	border-radius: 2px;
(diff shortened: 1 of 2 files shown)


votdev:
Thanks for the idea, but I already decided against using chips/badges in the user/group table in OMV6 because it is too much waste of space.

Bildschirmfoto vom 2020-02-08 09-32-48

vs.

Bildschirmfoto vom 2020-02-08 09-31-44

Because of that i doesn't make sense to introduce it in OMV5 and revert it in OMV6. Another reason is that OMV5 has feature freeze, i don't want to add new features to the UI which is replaced soonish.

abo-khalid:
I totally agree it's a waste of space even without the tag styling.  If we have more than 10 users, i believethe page will look ugly and less organized..  
Another idea is to maintain line height row height and wrap the content. This will generate equally listed rows in a much cleaner waym Then, make the row expands on click or maybe popup a new window that shows user details...
Also, filter function for group in the table will be a great feature I guess it's already built in into Extjs.

Best

votdev:
> Another idea is to maintain line height row height and wrap the content. This will generate equally listed rows in a much cleaner waym Then, make the row expands on click or maybe popup a new window that shows user details...
> Also, filter function for group in the table will be a great feature I guess it's already built in into Extjs.

As already said, no new UI features because it is not worth by the fact it will be replaced in OMV6.

abo-khalid:
I was referring to OMV6

PR #632, opened 2020-03-02
Fix path to UPS tools in nut for Buster

kayhannay:
The path to the UPS tools has changed in Debian Buster. This commit fixes the path in the monit configuration.

### DIFF
--- a/deb/openmediavault-nut/srv/salt/omv/deploy/monit/services/files/nut.j2
+++ b/deb/openmediavault-nut/srv/salt/omv/deploy/monit/services/files/nut.j2
@@ -16,7 +16,7 @@ check process nut-monitor with matching upsmon
     noalert {{ email_config.primaryemail }}
 {%- endif %}
 
-check program nut-upsc-{{ nut_config.upsname }} with path "/usr/bin/upsc {{ nut_config.upsname }}"
+check program nut-upsc-{{ nut_config.upsname }} with path "/bin/upsc {{ nut_config.upsname }}"
     group nut
-    start program = "/usr/sbin/upsdrvctl start"
+    start program = "/sbin/upsdrvctl start"
     if status != 0 for 2 cycles then restart


ryecoaaron:
Actually, /usr/bin/upsc is the correct path.  You see it in /bin/ because it is a symlink to /usr/bin

votdev:
> Actually, /usr/bin/upsc is the correct path. You see it in /bin/ because it is a symlink to /usr/bin
> 
> $ ls -ld /bin
> lrwxrwxrwx 1 root root 7 Jan 13 07:36 /bin -> usr/bin

That's exacly what i want to write, too.

kayhannay:
On my system it is not a link:

$ls -ld /bin 
drwxr-xr-x 2 root root 12288 Mär  1 21:43 /bin

and it seems that nut-client creates files in /bin:

$dpkg -L nut-client
...
/bin/upsc
...

ryecoaaron:
I think this is probably done for older systems that have been upgraded and don't have the /bin to /usr/bin symlink.  I would add the symlink myself since it is the standard across many Linux distros now.

votdev:
The PR has been closed because /usr/bin is the default dierctory for binaries. 
```
Fedora (and other distributions) have finished work on getting rid of the separation of /bin and /usr/bin, as well as /sbin and /usr/sbin, /lib and /usr/lib, and /lib64 and /usr/lib64. All files from the directories in / will be merged into their respective counterparts in /usr, and symlinks for the old directories will be created instead.
```
See https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge/.

kayhannay:
This was introduced with Debian 10 and my system has been upgraded from Debian 9. So all people doing an upgrade will probably have this issue. But anyway, we can close this.

votdev:
OMV5 will be the first (and hopefully last) release without an official migration path because of the huge changes.

votdev:
> This was introduced with Debian 10 and my system has been upgraded from Debian 9. So all people doing an upgrade will probably have this issue. But anyway, we can close this.

IMO this must be migrated by the base OS (Debian) and not OMV.

kayhannay:
There is a tool available in Debian which merges the paths. It is called 'usrmerge' and handles the merge during installation. With OMV-nut installed, there is one conflict for /usr/bin/upssched-cmd, which is installed by nut-client as well. This could lead to problems, I think.

Setting the paths to /bin/ as in my merge request would keep the compatibility with upgraded systems. And nut-client installs its binaries in /bin anyway. So this package is not yet using the new merged directory structure (even though it works, because there is a link).

This just to share the information. My issue is fixed now and I don't resist on merging this request.

PR #662, opened 2020-04-02
fix some misspells and add new in Polish Translation

MPrusek:
fix some misspells and add new in Polish Translation
Signed-off-by: Michał Prusek <spam@domel.be>


votdev:
Thanks you for the contribution. Would you please so kind to create an account on Transifex (https://translate.openmediavault.org) to make the changes there. For small changes i would do it myself, but the changes are to huge to do it manually.

Issue #689, opened 2020-04-28
Network MTU value doesn't get applied


getSurreal:
Per the documentation for netplan you must match the nic by MAC address for the MTU value to be applied.  https://netplan.io/reference

### Steps to reproduce issue
Set the MTU to 9000 from the admin panel
Run ifconfig from the termianl and the MTU is still 1500

### Resolution
Edit /etc/netplan/20-openmediavault-{your-nic}.yaml and add "match: macaddress"

Before:
```
network:
  ethernets:
    enp2s0:
      addresses:
      - 192.168.0.12/24
      gateway4: 192.168.0.1
      dhcp4: false
      dhcp6: false
      link-local: []
      nameservers:
        addresses:
        - 192.168.0.2
      mtu: 9000
```

After:

```
network:
  ethernets:
    enp2s0:
      match:
        macaddress: 00:25:90:a3:ff:c0
      addresses:
      - 192.168.0.12/24
      gateway4: 192.168.0.1
      dhcp4: false
      dhcp6: false
      link-local: []
      nameservers:
        addresses:
        - 192.168.0.2
      mtu: 9000
```

votdev:
You are right, but in this case I had to remove this option from the UI and database because OMV does not store the MAC address. Thx for the hint.

getSurreal:
The mac address is available from the salt grains.  I've submitted a PR that add that to the jinja template.

PR #690, opened 2020-04-28
Add macaddress to support custom mtu

getSurreal:
The netplan yaml file requires "match: macaddress" in order to support custom mtu values.

Fixes: https://github.com/openmediavault/openmediavault/issues/689

Signed-off-by: James LeRoy <james@getsurreal.com>


getSurreal:
I did some further testing with bonding and vlans and realized the mac address is not needed in those.  I did not test wifi though.

votdev:
Thx

votdev:
Had to undo the changes for wifi interfaces because of `ERROR: wlx0014785193b8: networkd backend does not support wifi with match:, only by interface name`. See https://github.com/openmediavault/openmediavault/commit/8fef89e2a0ffddef56ce099c18ca5e346298c266.

PR #729, opened 2020-05-31
fix RPC enumerateMountedFilesystems optional parameters

jlayec:
update array_boolval function in /usr/share/php/openmediavault/functions.inc
* default value with a 3rd optional parameter
* work also without parameter

* default value with a 3rd optional parameter
[/usr/share/openmediavault/engined/rpc/filesystemmgmt.inc](https://github.com/openmediavault/openmediavault/blob/master/deb/openmediavault/usr/share/openmediavault/engined/rpc/filesystemmgmt.inc) is calling array_boolval with 3 parameters, the last one is the default value not used by the function -> this has been fixed

* work also without parameter
omv-rpc 'FileSystemMgmt' 'enumerateMountedFilesystems'
is not working
PHP Fatal error: Uncaught TypeError: Argument 1 passed to array_boolval() must be of the type array, null given, called in /usr/share/openmediavault/engined/rpc/filesystemmgmt.inc on line 213 and defined in /usr/share/php/openmediavault/functions.inc:239
-> this has been fixed

- [ ] References issue

### DIFF
--- a/deb/openmediavault/usr/share/php/openmediavault/functions.inc
+++ b/deb/openmediavault/usr/share/php/openmediavault/functions.inc
@@ -234,11 +234,12 @@ function array_rand_value(array $array, $num = 1) {
  * @ingroup api
  * @param array $array An array with keys.
  * @param string $key The key of the element.
+ * @param string $default The default value returned if not found.
  * @return Returns the boolean value of the given key.
  */
-function array_boolval(array $array, $key) {
+function array_boolval($array, $key, $default = false) {
 	if (!isset($array[$key]))
-		return false;
+		return $default;
 	return boolvalEx($array[$key]);
 }
 
--- a/deb/openmediavault/usr/share/php/openmediavault/functions.inc
+++ b/deb/openmediavault/usr/share/php/openmediavault/functions.inc
@@ -238,7 +238,9 @@ function array_rand_value(array $array, $num = 1) {
  * @return Returns the boolean value of the given key.
  */
 function array_boolval($array, $key, $default = false) {
-	if (!isset($array[$key]))
+	if (!is_array($array) || !isset($array[$key])) {
+		return $default;
+	}
 		return $default;
 	return boolvalEx($array[$key]);
 }
--- a/deb/openmediavault/usr/share/php/openmediavault/functions.inc
+++ b/deb/openmediavault/usr/share/php/openmediavault/functions.inc
@@ -238,7 +238,7 @@ function array_rand_value(array $array, $num = 1) {
  * @return Returns the boolean value of the given key.
  */
 function array_boolval($array, $key, $default = false) {
-	if (!is_array($array) || !isset($array[$key])) {
+	if (!is_array($array) || !array_key_exists($key, $array)) {
 		return $default;
 	}
 	return boolvalEx($array[$key]);
--- a/deb/openmediavault/usr/share/php/openmediavault/functions.inc
+++ b/deb/openmediavault/usr/share/php/openmediavault/functions.inc
@@ -241,7 +241,6 @@ function array_boolval($array, $key, $default = false) {
 	if (!is_array($array) || !isset($array[$key])) {
 		return $default;
 	}
-		return $default;
 	return boolvalEx($array[$key]);
 }
 


deb/openmediavault/usr/share/php/openmediavault/functions.inc
```diff
@@ -234,11 +234,12 @@ function array_rand_value(array $array, $num = 1) {
  * @ingroup api
  * @param array $array An array with keys.
  * @param string $key The key of the element.
+ * @param string $default The default value returned if not found.
  * @return Returns the boolean value of the given key.
  */
-function array_boolval(array $array, $key) {
+function array_boolval($array, $key, $default = false) {
```
votdev, line 240: Any reason why you removed the type hint?

votdev:
Nice spotting that `array_boolval` is called with 3 parameters.

> Argument 1 passed to array_boolval() must be of the type array, null given

The more interesting question is why the argument `$params` is `NULL`.

deb/openmediavault/usr/share/php/openmediavault/functions.inc
jlayec: try that command

`omv-rpc 'FileSystemMgmt' 'enumerateMountedFilesystems'`

you will get that error

```
May 31 16:39:30 proliant omv-engined[31288]: PHP Fatal error: Uncaught TypeError: Argument 1 passed to array_boolval() must be of the type array, null given, called in /usr/share/openmediavault/engined/rpc/filesystemmgmt.inc on line 213 and defined in /usr/share/php/openmediavault/functions.inc:239
May 31 16:39:30 proliant omv-engined[31288]: Stack trace:
May 31 16:39:30 proliant omv-engined[31288]: #0 /usr/share/openmediavault/engined/rpc/filesystemmgmt.inc(213): array_boolval(NULL, 'includeroot', false)
May 31 16:39:30 proliant omv-engined[31288]: #1 [internal function]: Engined\Rpc\OMVRpcServiceFileSystemMgmt->enumerateMountedFilesystems(NULL, Array)
May 31 16:39:30 proliant omv-engined[31288]: #2 /usr/share/php/openmediavault/rpc/serviceabstract.inc(123): call_user_func_array(Array, Array)
May 31 16:39:30 proliant omv-engined[31288]: #3 /usr/share/php/openmediavault/rpc/rpc.inc(86): OMV\Rpc\ServiceAbstract->callMethod('enumerateMounte...', NULL, Array)
May 31 16:39:30 proliant omv-engined[31288]: #4 /usr/sbin/omv-engined(537): OMV\Rpc\Rpc::call('FileSystemMgmt', 'enumerateMounte...', NULL, Array, 1)
May 31 16:39:30 proliant omv-engined[31288]: #5 {main}
```

in fact if array_boolval is called without params, $params get the NULL value and as the type expected is array in the function, we have the exception raised above.

The easiest way to remove the bug is to remove type in the expected parameters of that function.
And you did that for the $key, so I did it for $params. :)

deb/openmediavault/usr/share/php/openmediavault/functions.inc
```diff
@@ -234,11 +234,12 @@ function array_rand_value(array $array, $num = 1) {
...
 	if (!isset($array[$key]))
```
votdev, line 241: Could you please add this to your PR?
```suggestion
	if (!is_array($array) || !isset($array[$key])) {
		return $default;
	}
```

jlayec:
reviewing the entire file I discovered that there is `array_value` function similar

* and inspiring the way `array_value` function was implemented we could do it the same way

Proposed new way

```
function array_boolval($array, $key, $default = FALSE) {
	if (!is_array($array) || !array_key_exists($key, $array))
		return $default;
	return boolvalEx($array[$key]);
}
```

* maybe we could have the same problem with `array_value` function where array is an expected type
==> we should remove the array type.
==> and we should simplify the last return (as the isset is confirmed by array_key_exists). no?

```
function array_value($array, $key, $default = NULL) {
	if (!is_array($array) || !array_key_exists($key, $array))
		return $default;
	return $array[$key];
}
```

deb/openmediavault/usr/share/php/openmediavault/functions.inc
```diff
@@ -234,11 +234,13 @@ function array_rand_value(array $array, $num = 1) {
...
  * @param array $array An array with keys.
  * @param string $key The key of the element.
+ * @param string $default The default value returned if not found.
  * @return Returns the boolean value of the given key.
  */
-function array_boolval(array $array, $key) {
-	if (!isset($array[$key]))
-		return false;
+function array_boolval($array, $key, $default = false) {
+	if (!is_array($array) || !isset($array[$key])) {
```
votdev, line 241:
```suggestion
	if (!is_array($array) || !array_key_exists($key, $array)) {
```

Issue #730, opened 2020-06-06
Allow chrony to serve clients


syphr42:
### Description of issue/question
This is a feature request to allow chrony to serve clients.

Chrony is used to sync time, but it is not possible in OMV to configure chrony to allow clients. For chrony, this is done by adding the `allow` directive in the conf file.

Chrony configuration directives can be seen here: https://chrony.tuxfamily.org/doc/3.4/chrony.conf.html

### Steps to reproduce issue
N/A

### Versions report
N/A

I will submit a PR to add this feature.

PR #731, opened 2020-06-06
Allow chrony to serve clients

syphr42:
This change provides a way for users to specify 'allow' directives to chrony so that OMV can serve time requests. When "Use NTP server" is selected, "Allow clients" becomes available. This is a domain or IP list like "Time servers," however entering a partial IP address (e.g. '10.0') will allow all clients in unspecified subnets. See chrony docs for more information.

Fixes: #730

Signed-off-by: Gregory Moyer <moyerg@syphr.com>

- [x] References issue

### DIFF
--- a/deb/openmediavault/srv/salt/omv/deploy/chrony/files/etc-chrony-chrony_conf.j2
+++ b/deb/openmediavault/srv/salt/omv/deploy/chrony/files/etc-chrony-chrony_conf.j2
@@ -7,7 +7,7 @@
 
 {{ pillar['headers']['multiline'] }}
 
-{%- for timeserver in timeservers.replace(',', ' ').replace(';', ' ').split() %}
+{%- for timeserver in ntp_config.timeservers.replace(',', ' ').replace(';', ' ').split() %}
 server {{ timeserver }} iburst
 {%- endfor %}
 
@@ -39,3 +39,8 @@ rtcsync
 # Step the system clock instead of slewing it if the adjustment is larger than
 # one second, but only in the first three clock updates.
 makestep {{ makestep }}
+
+# Allow clients to request time with allow directives below.
+{%- for client in ntp_config.clients.replace(',', ' ').replace(';', ' ').split() %}
+allow {{ client }}
+{%- endfor %}
--- a/deb/openmediavault/var/www/openmediavault/js/omv/module/admin/system/time/Time.js
+++ b/deb/openmediavault/var/www/openmediavault/js/omv/module/admin/system/time/Time.js
@@ -47,7 +47,8 @@ Ext.define("OMV.module.admin.system.time.Time", {
 				"hour",
 				"minute",
 				"second",
-				"ntptimeservers"
+				"ntptimeservers",
+				"ntpclients"
 			],
 			conditions: [
 				{ name: "ntpenable", value: true }
@@ -61,6 +62,12 @@ Ext.define("OMV.module.admin.system.time.Time", {
 			properties: "disabled"
 		},{
 			name: "ntptimeservers",
+			conditions: [
+				{ name: "ntpenable", value: true }
+			],
+			properties: [ "allowBlank", "!readOnly" ]
+		},{
+			name: "ntpclients",
 			conditions: [
 				{ name: "ntpenable", value: true }
 			],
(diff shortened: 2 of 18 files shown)


deb/openmediavault/srv/salt/omv/deploy/chrony/files/etc-chrony-chrony_conf.j2
```diff
@@ -39,3 +39,8 @@ rtcsync
 # Step the system clock instead of slewing it if the adjustment is larger than
 # one second, but only in the first three clock updates.
 makestep {{ makestep }}
+
+# Allow clients to request time with allow directives below
```
votdev, line 43:
```suggestion
# Allow clients to request time with allow directives below.
```

deb/openmediavault/var/www/openmediavault/js/omv/module/admin/system/time/Time.js
```diff
@@ -141,6 +148,14 @@ Ext.define("OMV.module.admin.system.time.Time", {
 				allowBlank: true,
 				readOnly: true,
 				value: "pool.ntp.org"
+			},{
+				xtype: "textfield",
+				name: "ntpclients",
+				fieldLabel: _("Allow clients"),
```
votdev, line 154: IMO the label does not reflect the intention of this configuration, the current name suggests a boolean value.

I would also like to see a hint which describes the intention of this field.

deb/openmediavault/srv/salt/omv/deploy/chrony/default.sls
```diff
@@ -27,6 +27,7 @@ configure_chrony:
     - template: jinja
     - context:
         timeservers: "{{ config.ntp.timeservers }}"
+        clients: "{{ config.ntp.clients }}"
```
votdev, line 30: Due the fact that we now pass multiple configuration options to the template, i suggest to use this.
```suggestion
    - context:
        ntp_config: "{{ config.ntp }}"
```

syphr42:
@votdev Thanks for the comments. I'll work on the changes. Specifically for the database migration script, should I edit the latest one or create a new one?

votdev:
> @votdev Thanks for the comments. I'll work on the changes. Specifically for the database migration script, should I edit the latest one or create a new one?

Please add `Issue #730: Allow chrony to serve clients` to the changelog for the version `5.5.3` i currently commited. Then use `5.5.3` for the migration script.

deb/openmediavault/srv/salt/omv/deploy/chrony/default.sls
syphr42: I'm having a bit of trouble with this one. When I change the context, the Jinja 2 template fails on the references to timeservers and clients. I tried a few different ways of referencing the variables inside the template, but they've all failed. My Python is a little rusty - can you give me a hint as to the template change that is needed to go with this?

syphr42:
I found that when I would put a value like `10.0` in the allowed clients field, it would cause validation issues in some cases because Python was interpreting the value as a number and then the type validation against `str` would fail. I didn't see an good way around this, so I changed the format to CIDR which is also allowed by Chrony.

deb/openmediavault/var/www/openmediavault/js/omv/module/admin/system/time/Time.js
syphr42: I updated this to "Allowed clients" and added some hint text.

votdev:
Could you finally squash all commints into one. This will make it easier to cherry-pick it in the OMV5 branch.

deb/openmediavault/srv/salt/omv/deploy/chrony/default.sls
votdev: oh right, it must be

```
    - context:
        ntp_config: {{ config.ntp | json }}
```
You need to replace `timeservers` and `clients` with `ntp_config.timeservers` and `ntp_config.clients` in the `etc-chrony-chrony_conf.j2` file.

syphr42: This fails with

```
Rendering SLS 'base:omv.deploy.chrony.default' failed: while parsing a block mapping in "<unicode string>", line 29, column 9 did not find expected key in "<unicode string>", line 29, column 24
```

I'll investigate. My lack of experience with Python is showing. :grimacing:

syphr42: I got it. Had quotes around `{{ config.ntp | json }}` leftover from before that was throwing everything off.

syphr42:
This PR should be good to go now. Let me know if you spot anything else that needs attention.

votdev:
Thx for the improvement. 

Superseded by https://github.com/openmediavault/openmediavault/pull/734.

Issue #648, opened 2020-03-18
UPS configuration fails to apply on Debian Buster


martgras:
### Steps to reproduce issue

Enable remote monitoring ( or any other change in the UPS config ) and apply the changes 

````
OMV\ExecException: Failed to execute command 'export PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin; export LANG=C.UTF-8; omv-salt deploy run monit 2>&1' with exit code '1':

        ID: test_monit_config
    Function: cmd.run
        Name: monit -t
      Result: False
     Comment: Command "monit -t" run
     Started: 14:28:47.336349
    Duration: 17.971 ms
     Changes:
              ----------
              pid:
                  13169
              retcode:
                  1
              stderr:
                  /etc/monit/conf.d/openmediavault-nut.conf:13: Program does not exist: '/usr/bin/upsc'
                  /etc/monit/conf.d/openmediavault-nut.conf:13: Program does not exist: 'tvups'
                  /etc/monit/conf.d/openmediavault-nut.conf:15: Program does not exist: '/usr/sbin/upsdrvctl'
                  AssertException: File '/usr/bin/upsc' does not exist
                   raised in Command_new at src/system/Command.c:359
              stdout:

````
The reason is that on Debian Buster upsc is in /bin and not in /usr/bin  and upsdrvctl  is in /sbin

changing  /srv/salt/omv/deploy/monit/services/files/nut.j2  fixed it for me :

````
...
check program nut-upsc-{{ nut_config.upsname }} with path "/bin/upsc {{ nut_config.upsname }}"
    group nut
    start program = "/sbin/upsdrvctl start"
    if status != 0 for 2 cycles then restart
````

Ref : https://packages.debian.org/buster/amd64/nut-client/filelist and https://packages.debian.org/buster/amd64/nut-server/filelist

### Versions report
OMV 5.3.8-1

votdev:
https://github.com/openmediavault/openmediavault/pull/632

PR #750, opened 2020-07-01
Update nut.j2 /bin and /sbin locations

devn3t:
changed location of upsc and upsdrvctl commands as they are no longer located at /usr/bin and /usr/sbin accordingly.  Should be located at /sbin and /bin.

- [ ] References issue

### DIFF
--- a/deb/openmediavault-nut/srv/salt/omv/deploy/monit/services/files/nut.j2
+++ b/deb/openmediavault-nut/srv/salt/omv/deploy/monit/services/files/nut.j2
@@ -18,8 +18,8 @@ check process nut-monitor with matching upsmon
 {%- endif %}
 
 {%- if nut_config.mode != 'netclient' %}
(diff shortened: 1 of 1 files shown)


devn3t:
Problem found here:  https://forum.openmediavault.org/index.php?thread/31458-unable-to-use-network-ups-tools-to-monitor-my-ups/

Reproduced on my local system.  Fix resolves issue locally on my system as well (5.5.3-1 Usal)

votdev:
The issue has been closed because `/usr/bin` is the default directory for binaries. See https://freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge.

Duplicate of https://github.com/openmediavault/openmediavault/pull/632 and https://github.com/openmediavault/openmediavault/issues/648.

devn3t:
Thanks!  Didn't know it existed!

Issue #763, opened 2020-07-16
Support ejecting tapes in usbbackup plugin.


votdev:
The `eject` command can be used to unmount the filesystem after a successful rsync job. A nice side-effect is that tape devices will be ejected automatically.

See https://forum.openmediavault.org/index.php?thread/33487-usbbackup-eject/.

### Steps to reproduce issue
(Include debug logs if possible and relevant.)

PR #764, opened 2020-07-16
Issue #763: Support ejecting tapes.

votdev:
After a successful sync in non-interactive mode use the 'eject' command to unmount the filesystem, making the device node disappear and activate the loading mechanism for devices like tape drives.

Fixes: https://github.com/openmediavault/openmediavault/issues/763

- [x] References issue


votdev:
It seems there are some problems with this PR. It is not possible to add additional jobs when there are existing active jobs. When you plug the device the jobs are executed and at the end the device is ejected, which means the device node is removed. Because of that the UI is not able to show the plugged USB device in the `Create` dialog which makes it impossible to add new jobs using that device.

PR #765, opened 2020-07-17
User can choose administrator users

julienblitte:
[Short description]
User can choose administrator users by placing them into the group openmediavault-webgui.

[A longer multiline description]
Not being able to change administrator login is a lack of security (login already known) as well as not being able to have several administrators on the same machine (password sharing).
Also not being able to customize the login for the administrator is a lack of comfort for the user.
Using a coded login is bad practice.
This pull request resolves all the above issues by detecting who is openmediavault web administrator. The user can choose who is administrator by placing the account into the group openmediavault-webgui.

This is retro-compatible with previous behavior ('admin' account still created with 'openmediavault' password by default).

Signed-off-by: [Julien Blitte] <[julien.blitte@gmail.com]>

### DIFF
--- a/deb/openmediavault/debian/openmediavault.postinst
+++ b/deb/openmediavault/debian/openmediavault.postinst
@@ -75,6 +75,11 @@ case "$1" in
 					--comment 'openmediavault WebGUI Administrator' \
 					admin
 			fi
+			####################################################################
+			# User 'admin' is added to group 'openmediavault-webgui' to provide
+			# access to web interface.
+			####################################################################
+			adduser admin openmediavault-webgui
 
 			####################################################################
 			# Create various required files.
--- a/deb/openmediavault/var/www/openmediavault/rpc/session.inc
+++ b/deb/openmediavault/var/www/openmediavault/rpc/session.inc
@@ -36,6 +36,23 @@ class OMVRpcServiceSession extends \OMV\Rpc\ServiceAbstract {
 	}
 
 	/**
+	* Indicate if user is an administrator (member of group openmediavault-webgui)
+	* @param username The name of the user.
+	* @return TRUE is user is administrator, FALSE otherwise.
+	*/
+	private function isAdmin($username) {
+		$group = posix_getgrnam('openmediavault-webgui');
+		$adminList = $group ? $group['members'] : array();
+		
+		if (empty($adminList)) {
+			syslog(LOG_WARNING, 'No openmediavault administrator could be find in the group openmediavault-webgui');
+			// failback to name admin account
+			$adminList = array('admin');
+		}
+		return in_array($username, $adminList);
+	}
+
+	/**
 	 * Login user.
 	 * @param params The method parameters containing the following fields:
 	 *   \em username The name of the user.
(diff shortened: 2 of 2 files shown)


deb/openmediavault/debian/openmediavault.postinst
```diff
@@ -75,6 +75,11 @@ case "$1" in
 					--comment 'openmediavault WebGUI Administrator' \
 					admin
 			fi
+			####################################################################
+			# User 'admin' is added to group 'openmediavault-webgui' to provide
+			# access to web interface.
+			####################################################################
+			adduser admin openmediavault-webgui
```
votdev, line 82: This user is for PHP-FPM only.

deb/openmediavault/var/www/openmediavault/rpc/session.inc
```diff
@@ -35,6 +35,23 @@ class OMVRpcServiceSession extends \OMV\Rpc\ServiceAbstract {
 		$this->registerMethod("logout");
 	}
 
+	/**
+	* Indicate if user is an administrator (member of group openmediavault-webgui)
+	* @param username The name of the user.
+	* @return TRUE is user is administrator, FALSE otherwise.
+	*/
+	private function isAdmin($username) {
```
votdev, line 43: Try to avoid using the product name in any message (because of branding issues).

I'm asking myself why it should be necessary to test the result of `posix_getgrnam`? I've not tested the code, but are you sure this works here? `nginx/php-fpm` does not run with the correct privileges to access the user database, this must be done in `omv-engined` which has the correct privileges.

The whole function can be simplified (under the assumption `nginx/php-fpm` has the right privileges):
```
    $group = posix_getgrnam("openmediavault-webgui");
    $adminList = $group ? $group['members'] : [ "admin" ];
    return in_array($username, $adminList);
```

votdev:
I will take over this PR to implement that feature correctly. Thanks for the initial idea.

deb/openmediavault/debian/openmediavault.postinst
julienblitte: Indeed, but I wanted to avoid to create another group as we have already 3 new groups for OMV.

deb/openmediavault/var/www/openmediavault/rpc/session.inc
julienblitte: Just for information, on my system, `posix_getgrnam` and this changes works well (it has been tested before I pushed it) as the group reading access is for accessible to everyone.
I have implemented the simpliest way but maybe not the cleanest, you would do better than me for this :)

julienblitte:
Thanks. Waiting for correct implementation of this new feature when you have the time to do it.
Any help I might be able to provide (tests, etc.), feel free to ask.

votdev:
The following commits are required to get it working in OMV5. Had to backport them from OMV6.

136dbc7d3dafbea7e632cd4237b34ec17b473f04
551ba191d3c17ee11b3e433d50a7bc2abb8354a1
eff97fded5f3c5042af7c8660f31fc54c664d096
3feaed824b7ddd27fa3fde2a461f3b9c57c6b02a

PR #790, opened 2020-08-19
Also consider read-only users and groups in proftpd limits

### LINKED ISSUE #789: Read-only users or groups are not used in proftp config
### Description of issue/question

I believe there's an issues with both ifs [here](https://github.com/openmediavault/openmediavault/blob/aa329e2547cd392b28f5f070ca466fb367ea42b6/deb/openmediavault/srv/salt/omv/deploy/proftpd/modules/files/mod_core.j2#L142-L147) as it will only work if you have at least one user or group with perms 7, not if your list only consists of perms 5 (which is my case). I believe those ifs should concatenate the lists before checking the length.

### Steps to reproduce issue
I believe you just need to only include a single group with read-only permissions on a share.

### Versions report
5.5.8-1

Marsup:
Fixes #789.

- [X] References issue

I'm not sure how to test that, so I went for a naive fix as I'm not very familiar with j2.
I might add that I tested it on my own setup and it seems to work properly.

### DIFF
--- a/deb/openmediavault/srv/salt/omv/deploy/proftpd/modules/files/mod_core.j2
+++ b/deb/openmediavault/srv/salt/omv/deploy/proftpd/modules/files/mod_core.j2
@@ -139,10 +139,10 @@ DisplayLogin {{ display_login }}
     DenyAll
   </Limit>
   <Limit READ DIRS>
-{%- if limitall_allowuser | length > 0 %}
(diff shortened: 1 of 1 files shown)


votdev:
https://github.com/openmediavault/openmediavault/commit/6c91b2b25a68aa6c85cad3aa8a4e1132ec4740ec will update the changelog and mark the `proftpd` module as dirty.

PR #806, opened 2020-09-05
Remove unnecessary line break in CPU usage legend

sebeksd:
When using OMV_COLLECTD_RRDTOOL_GRAPH_WIDTH to make graph wider CPU legend was still split into two lines even when there was enough space for one line

Before (width 800 and 200):
image
image

After (width 800 and 200):
image
image

Signed-off-by: sebeksd <18705953+sebeksd@users.noreply.github.com>

### DIFF
--- a/deb/openmediavault/usr/share/openmediavault/mkrrdgraph/plugins.d/cpu.py
+++ b/deb/openmediavault/usr/share/openmediavault/mkrrdgraph/plugins.d/cpu.py
@@ -63,9 +63,10 @@ class Plugin(openmediavault.mkrrdgraph.IPlugin):
         args.append('AREA:waitio{color_cpu_waitio}:"Wait-IO":STACK'.format(**config))
         args.append('AREA:nice{color_cpu_nice}:"Nice":STACK'.format(**config))
         args.append('AREA:user{color_cpu_user}:"User":STACK'.format(**config))
-        args.append('AREA:softirq{color_cpu_softirq}:"Soft-IRQ\c":STACK'.format(**config))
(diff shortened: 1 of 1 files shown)


votdev:
Please adapt your PR according to https://openmediavault.readthedocs.io/en/5.x/development/contribute.html#contribution-rules.

votdev:
I'm sorry, the `Signed-off-by` must be added to the commit message, otherwise this information can get lost.

sebeksd:
Signed-off-by in both commit and pull request

votdev:
Seems it does not work. The first and last legend entries are cut off.
Bildschirmfoto vom 2020-09-05 22-23-53

votdev (changes requested):
Seems it does not work with default values. The first and last legend entries are cut off.
Bildschirmfoto vom 2020-09-05 22-23-53

sebeksd:
Fixed. (Amended last commit)

It's my first time changing something in rrdtool graphs, but it seems that old graph had same issue as you pointed (but of course not on default settings).
Now after a little change it seems to play well with every width (not cutting anything), only thing that I don't like in my change is regarding default width of 400 because with that settings single "label" is in new line, but maybe this is fine ?
Changing default to 450 will make it a single line (but not sure if you like that).

Some screens:
at 400 (default)
image

at 450
image

at 300
image

at 800
image

votdev (changes requested):
Thanks, looks OK now. Could you please squash your PR into one commit? The commit `Merge branch 'master' into master` isn't really necessary.

sebeksd:
Merge commit deleted, all changes are in single commit, is it ok now ?

Issue #665, opened 2020-04-04
Disks with identical attributes to others not being enumerated


Catdaemon:
### Description of issue/question
I have a drive array consisting of two identical 4-bay USB3 JBOD enclosures. These are running off a raspberry pi 4. I have successfully set this up with 7 drives. Upon adding the 8th drive, all of the drives in that enclosure disappear from the Disks section of OMV. The drives in the other enclosure are still there. Everything else works fine and they are all detected by the operating system. All of the filesystems on the disks show in the File Systems area, but I am obviously unable to create a filesystem on the new drive via the UI.  If I then remove the added drive they all show in the Disks section again.

I haven't read the code, but I have a feeling this may be caused by the cheap-ish JBOD enclosures I'm using, they're JMicron controllers with identical IDs, serial numbers etc. which is annoying and I assume is the problem. I thought this might be similar to #639 maybe except my filesystems work fine, and all was good until the last disk was added. Here they are via `ls -l /dev/disk/by-id/ | awk '{print $9,$10,$11}'`:
```
usb-JMicron_Generic_DISK00_0123456789ABCDEF-0:0 -> ../../sde
usb-JMicron_Generic_DISK00_0123456789ABCDEF-0:0-part1 -> ../../sde1
usb-JMicron_Generic_DISK01_0123456789ABCDEF-0:1 -> ../../sdf
usb-JMicron_Generic_DISK01_0123456789ABCDEF-0:1-part1 -> ../../sdf1
usb-JMicron_Generic_DISK02_0123456789ABCDEF-0:2 -> ../../sdg
usb-JMicron_Generic_DISK02_0123456789ABCDEF-0:2-part1 -> ../../sdg1
usb-JMicron_Generic_DISK03_0123456789ABCDEF-0:3 -> ../../sdh
usb-JMicron_Generic_DISK03_0123456789ABCDEF-0:3-part1 -> ../../sdh1
```
I added an entry to my udev rules to stop these links being overwritten by the duplicates and see if this would fix it:
```
# JMicron drive hack
KERNEL=="sd*", ENV{ID_VENDOR}=="JMicron", PROGRAM="/root/get_disk_serial.sh %k", SYMLINK+="disk/by-id/usb-enc-%c"
```

This makes the drives appear as follows:
```
usb-enc-GTA000PAG91YEA -> ../../sdb1
usb-enc-GTA000PAGAEBHA -> ../../sdh1
[...]

PR #831, opened 2020-10-20
Update 99-openmediavault-dev-disk-by-id.rules

hannesd:
Hi,

I had the same problem described in https://github.com/openmediavault/openmediavault/issues/665. Because the USB device reported the same serial for two different USB hard drives the link under /dev/disks/by-id got overwritten by whichever drive was added last. In the front under Disks or SMART only one of the two drives was listed.

The already existing fix in /etc/udev/rules.d/99-openmediavault-dev-disk-by-id.rules does indeed correct the ID_SERIAL* env variables, but it will not fix the symlink and therefore the front end will only show one of the drives despite that rule.

I added the two SYMLINK lines. At first I added twice (used += instead of = twice). This, however, created a third symlink in .../by-id/. Two symlinks had the fixed ID_SERIAL and the third was the "old", faulty symlink. For reasons unclear to me it is that faulty symlink that gets picked up by openmediavault as the devicefile and while the command "omv-rpc DiskMgmt enumerateDevices" enumerates both drives correctly, the front end will still only show one drive (I assume that the front end does its own filtering based on the devicefile property).

So I ended up with the solution provided here: first overwrite any existing SYMLINKs and add the by-path and the fixed by-id symlinks.

However, I have a hunch that a better solution would be to rename the rules files to something < 60-...rules instead of 99-...rules. The file which creates the faulty symlink is /lib/udev/rules.d/60-persistent-storage.rules. If the ID_SERIAL field is fixed before that rules file is parsed the extra SMYLINK magic would not be required. Yet I cannot recommend this course of action as my knowledge of udev is very limited and I do not know what other unforeseen consequences this might have.

Cheers
-- Hannes

- [x] References issue

### DIFF
--- a/deb/openmediavault/etc/udev/rules.d/99-openmediavault-dev-disk-by-id.rules
+++ b/deb/openmediavault/etc/udev/rules.d/99-openmediavault-dev-disk-by-id.rules
@@ -31,7 +31,9 @@ ACTION=="add", KERNEL=="sd*", SUBSYSTEMS=="usb", \
   ENV{ID_VENDOR}=="JMicron", \
   PROGRAM="/usr/lib/udev/serial_id %N", \
   ENV{ID_SERIAL}="USB-%c-$env{ID_INSTANCE}", \
-  ENV{ID_SERIAL_SHORT}="%c"
+  ENV{ID_SERIAL_SHORT}="%c", \
+  SYMLINK="disk/by-path/$env{ID_PATH}", \
+  SYMLINK+="disk/by-id/$env{ID_BUS}-$env{ID_SERIAL}"
 
 # JMicron JMS561U two ports SATA 6Gb/s bridge
 # https://www.sabrent.com/product/EC-DFLT/usb-3-0-sata-external-hard-drive-lay-flat-docking-station-2-5-3-5in-hdd-ssd/


votdev:
I will think about renaming the file to something like `60-...`to let do udev the SYMLINK magic. In the meanwhile i will merge your PR. Thx for the contribution.

votdev:
Could you please adapt the commit message according https://openmediavault.readthedocs.io/en/5.x/development/contribute.html#contribution-rules.

votdev:
Could you please test https://github.com/openmediavault/openmediavault/pull/835? I think renaming the file is the better solution than doing the SYMLINK magic ourself. Thanks for the tip.

hannesd:
My pleasure :)

Seeing how #835 does not work for me I have updated the commit message and signed the CLA.

votdev:
Thank you.

votdev:
Is it possible that you test https://github.com/openmediavault/openmediavault/pull/843 and check if it still works?

hannesd:
> Is it possible that you test #843 and check if it still works? The UDEV rules need to be more restrictive, so i added `ENV{DEVTYPE}=="disk", \`.

Works for me (JMicron case). I've set up a fresh 5.5.16-1 (Usul) install and replaced the file (`99-openmediavault-dev-disk-by-id.rules`) with that from the pull request. After a restart the symlinks in `/dev/disk/by-id/` are correctly named by our custom rule and all drives are listed in the GUI.

votdev:
> > Is it possible that you test #843 and check if it still works? The UDEV rules need to be more restrictive, so i added `ENV{DEVTYPE}=="disk", \`.
> 
> Works for me (JMicron case). I've set up a fresh 5.5.16-1 (Usul) install and replaced the file (`99-openmediavault-dev-disk-by-id.rules`) with that from the pull request. After a restart the symlinks in `/dev/disk/by-id/` are correctly named by our custom rule and all drives are listed in the GUI.

Have you tested the version that starts with
```
ACTION=="remove", GOTO="omv_dev_disk_by_id_end"
KERNEL!="sd*", GOTO="omv_dev_disk_by_id_end"
SUBSYSTEM!="block", GOTO="omv_dev_disk_by_id_end"
ENV{DEVTYPE}!="disk", GOTO="omv_dev_disk_by_id_end"
```
I'm asking because i have forced push a newer version at the same time you posted this comment.

hannesd:
> I'm asking because i have forced push a newer version at the same time you posted this comment.

It appears I have taken the code after you used the force. My file starts with the lines you mention and I just diffed the current state of https://github.com/openmediavault/openmediavault/pull/843/files vs. what's on my machine and the only thing that has changed are some comments.

Issue #832, opened 2020-10-20
Notification issue: Mailgun mail address format in filter cause rejected mails (Former Name: Postfix - Header differences cause mail rejection of OMV notifications)


that-coding-human:
I tried to setup my mail notification using mailgun. I am using mailgun as it is a convenient mail relay for automatically sending server notifications through smtp as well as API, which makes it very flexible and a perfect match if you want ONE single point of contact for sending notifications from different parts of your infrastructure. (Just as an explanation why I am not using gmail :-) )

I setup the Notifications with the following settings:
SMTP Server: smtp.mailgun.org
SMTP Port: 465
Encryption Mode: SSL/TLS
Sender E-Mail: postmaster@sandbox1234567890.mailgun.org
Authentication required: true
Username: postmaster@sandbox1234567890.mailgun.org
Password: TottallyObviouslyNotMyPassword

Recipient:
myAdress@someDomain.com

For small setups and testing (before going productive) the free offering of mailgun is sufficient, but it requires to register and verify the recipient adresses. I did this, as the recipient for this setup is my own adress.

When pressing "send a test email", the following gets placed in the postfix qeue.
I used the command: `postcat -q 3E5B542110D `
#############################################################################

```
*** ENVELOPE RECORDS deferred/3/3E5B542110D ***
message_size:             470             353               2               0             470               0
message_arrival_time: Tue Oct 20 10:24:30 2020
create_time: Tue Oct 20 10:24:30 2020
named_attribute: rewrite_context=local
sender_fullname: root
sender: postmaster@sandbox1234567890.mailgun.org
named_attribute: notify_flags=1
original_recipient: openmediavault-notification@localhost.localdomain
done_recipient: openmediavault-notification@localhost.localdomain
*** MESSAGE CONTENTS deferred/3/3E5B542110D ***
Received: by myNAS.local (Postfix, from userid 0)
        id 3E5B542110D; Tue, 20 Oct 2020 10:24:30 +0200 (CEST)
To: myAdress@someDomain.com
Subject: Test message from myNAS.local
Date: Tue, 20 Oct 2020 10:24:30 +0200
From: postmaster@sandbox1234567890.mailgun.org
Message-Id: <20201020082430.3E5B542110D@myNAS.local>

This is an automatically generated mail message from the system notification service running on myNAS.local.
*** HEADER EXTRACTED deferred/3/3E5B542110D ***
named_attribute: dsn_orig_rcpt=rfc822;myAdress@someDomain.com
[...]

PR #834, opened 2020-10-21
Use PHP_EOL to separate email headers.

votdev:
According to

* https://richardwarrender.com/2007/09/problem-with-php-mail-and-additional-headers/
* https://serverfault.com/questions/325506/php-mail-headers-not-working-properly-with-postfix

it is better to use PHP_EOL ('\n' on Linux) to separate headers.

- [x] References issue

### DIFF
--- a/deb/openmediavault/usr/share/php/openmediavault/email.inc
+++ b/deb/openmediavault/usr/share/php/openmediavault/email.inc
@@ -64,7 +64,7 @@ class Email {
 		if (!empty($this->from))
 			$headers[] = sprintf('From: %s', $this->from);
 		return mail($this->to, $this->subject, $this->message,
-		  implode("\r\n", $headers));
+			implode(PHP_EOL, $headers));
 	}
 
 	/**


deb/openmediavault/usr/share/php/openmediavault/email.inc
```diff
@@ -64,7 +64,7 @@ class Email {
 		if (!empty($this->from))
 			$headers[] = sprintf('From: %s', $this->from);
 		return mail($this->to, $this->subject, $this->message,
-		  implode("\r\n", $headers));
+			implode(PHP_EOL, $headers));
```
that-coding-human, line 67: This change did not change the behaviour of the mails being rejected by the mail provider. Anyway, according to the information you provided, it is better to keep PHP_EOL at this point. I could not observe any change in the behaviour of how the mails get formatted or placed in the queue.

that-coding-human (commented):
It seems like this did not change the behaviour. As it is recommended to use PHP_EOL, this change can be kept and merged anyway, but maybe this branch should be kept open and unmerged, until  issue #832 is resolved

PR #835, opened 2020-10-21
Rename UDEV rules file

votdev:
Rename UDEV rules file `99-openmediavault-dev-disk-by-id.rules` to `01-openmediavault-fix-serial.rules`. It is used to fix the serial number problem of some USB PATA/SATA bridge controllers.

The new filename should ensure that the file is processed before `/lib/udev/rules.d/60-persistent-storage.rules`. This way the SYMLINK magic is done by this rules instead of overriding the SYMLINKS ourself.

- [ ] References issue

### DIFF
--- a/deb/openmediavault/debian/changelog
+++ b/deb/openmediavault/debian/changelog
@@ -1,6 +1,9 @@
 openmediavault (5.5.13-1) stable; urgency=low
 
   * Fix bug in quota deployment on XFS filesystems.
+  * Rename UDEV rules file 99-openmediavault-dev-disk-by-id.rules to
+    01-openmediavault-fix-serial.rules. It is used to fix the serial
+    number problem of some USB PATA/SATA bridge controllers.
   * Issue #829: Show the listing of directory entries to which the
     anonymous user normally has no access permissions.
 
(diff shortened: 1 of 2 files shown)


hannesd:
Unfortunately it doesn't work. Here is what I tried:

Applying the changes from this pull request (renaming and removing the SYMLINK lines) does not change udev's behavior for the better. While previously at least the ID_SERIAL got corrected, now even ID_SERIAL is wrong on top of the faulty symlink.

I dug into it a bit and identified this line in `/lib/udev/rules.d/60-persistent-storage.rules` as the culprit:
```
# Fall back usb_id for USB devices
KERNEL=="sd*[!0-9]|sr*", ENV{ID_SERIAL}!="?*", SUBSYSTEMS=="usb", IMPORT{builtin}="usb_id"
```

If openmediavault's own rules are manually inserted **before** this statement, nothing will happen. If they are inserted **after** that statement, all is fine and the rules work as intended.

I found some documentation in the original C sources of udev: https://github.com/systemd/systemd/blob/69f30d4321c3bc8fa27270e3fd36a7a5b4d0a89e/src/udev/udev-builtin-usb_id.c#L213

To me it looks like builtins do not respect rule 12 from https://wiki.debian.org/udev:
`earlier rules have precedence over later rules - so put your customizations early in the rules.d file list`

Editing `/lib/udev/rules.d/60-persistent-storage.rules` is not an option I believe as this file would be overwritten by updates. Also that's not how the system is supposed to work.

I guess that means we are stuck with doing the SYMLINK magic in a file that is loaded **after** `/lib/udev/rules.d/60-persistent-storage.rules`. Maybe adding a bit of documentation to that file will help other's in the future. Let me know if I can be of assistance.

Maybe talking to the debian folks is an option. If the rules file `60-persistent-storage.rules` were to be split right after the "builtin" line and another file with a sufficiently high number contains the stuff that follows the "builtin" line, openmediavault could insert its own rules in between.

votdev:
>  Maybe adding a bit of documentation to that file will help other's in the future. Let me know if I can be of assistance.

Yes please, feel free to add some comments. I think it is better when someone else is doing that because you might have a different view on it than myself. Please open a new PR for that. Thanks.

PR #836, opened 2020-10-24
Whenever percent progress is refreshed split it into new line

sebeksd:
WebGUI multiline progress

Window is unable to refresh progress in single line (probably by design) but adding every new "refresh" at the end makes text hard to read,
spliting it into new lines makes it readable again and allow for "full command history" like before

Signed-off-by: sebeksd 18705953+sebeksd@users.noreply.github.com

image


votdev:
Thanks for the contribution, but i think converting the text in the backend is the better approach. So all consumers of the RPC will benefit from this improvement, not only the WebUI.

sebeksd:
You probably right. I didn't see this problem anywhere else (in emails it was fine) so I changed that only in WebUI 👍

PR #863, opened 2020-11-24
Fix string formatting for Repository

cjhutchi:
- [ ] References issue


votdev:
Wow, that's an old bug. Many thanks.

PrplHaz4:
> Wow, that's an old bug. Many thanks.

For some reason it is displaying fine on 5.5.16-1 and seems to have appeared for 5.5.17-x

votdev:
> > Wow, that's an old bug. Many thanks.
> 
> For some reason it is displaying fine on 5.5.16-1 and seems to have appeared for 5.5.17-x

Oh yes, you're right. It's new. Can't remember everything :-)

PR #890, opened 2020-12-19
update nut path for latest package

wheder:
The path to executables in Debian `nut-client_2.7.4-8_amd64.deb` and in Ubuntu `nut-client_2.7.4-9ubuntu1_amd64.deb` is different than is expected in this template.

Fixes [https://forum.openmediavault.org/index.php?thread/33333-omv5-nut-plugin-cannot-be-enabled-in-webui/&postID=254283#post254283]


wheder:
oh, my bad, there is a `usrmerge` command which should have been run with debian migration

Issue #906, opened 2020-12-31
UDEV rules to workaround invalid serial numbers causes problems


votdev:
**Describe the bug**
The UDEV rules to workaround a JMicron issue with USB enclosures where serial ids duplicated does not work as expected. According to https://github.com/systemd/systemd/issues/18111 the rules must be executed before `60-persistent-storage.rules` to do not cause problems with filesystems.

The fix is to adapt the existing rules by removing the `SYMLINK="..."` part and rename it to `59-xxx.rules` to be executed before `60-persistent-storage.rules`. The openmediavault UDEV rules will then modify the `ID_SERIAL` and `ID_SERIAL_SHORT` properties of the affected devices and `60-persistent-storage.rules` will take care about creating the device files.

This fix will sadly affect all users that are using JMicron based enclosures and that have applied the old openmediavault UDEV rules in the past. Such installations will break and need to be reconfigured from scratch related to filesystems and shared folders. The problem is that the filesystem configuration contains invalid device files after the new UDEV rules are applied.

votdev:
Closed because `60-persistent-storage.rules` will override the modified `ID_SERIAL` and `ID_SERIAL_SHORT` properties. Because of that `scsi_id` will export falsy serial id and override the correct properties again. Because of that we need to go the old way.

https://github.com/openmediavault/openmediavault/pull/909 will go the old way, but will improve debugging and the rules are applied earlier in the hope to prevent problems.

PR #907, opened 2020-12-31
UDEV rules to workaround invalid serial numbers causes problems

votdev:
The UDEV rules to workaround a JMicron issue with USB enclosures where serial ids duplicated does not work as expected. According to systemd/systemd#18111 the rules must be executed before 60-persistent-storage.rules to do not cause problems with filesystems.

The fix is to adapt the existing rules by removing the SYMLINK="..." part and rename it to 59-xxx.rules to be executed before 60-persistent-storage.rules. The openmediavault UDEV rules will then modify the ID_SERIAL and ID_SERIAL_SHORT properties of the affected devices and 60-persistent-storage.rules will take care about creating the device files.

This fix will affect some users that are using JMicron based enclosures or adaptors (e.g. Orangepi PC2) and that have applied the old openmediavault UDEV rules in the past. Such installations will break and need to be reconfigured from scratch related to filesystems and shared folders. The problem is that the filesystem configuration contains invalid device files after the new UDEV rules are applied.

Fixes: https://github.com/openmediavault/openmediavault/issues/906

- [x] References issue

### DIFF
--- a/deb/openmediavault/etc/udev/rules.d/59-openmediavault-serial-id.rules
+++ b/deb/openmediavault/etc/udev/rules.d/59-openmediavault-serial-id.rules
@@ -27,12 +27,22 @@
 # https://wiki.ubuntuusers.de/udev/
 # https://www.tecmint.com/udev-for-device-detection-management-in-linux/
 
-ACTION=="remove", GOTO="omv_dev_disk_by_id_end"
-KERNEL!="sd*", GOTO="omv_dev_disk_by_id_end"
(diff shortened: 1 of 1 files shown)


votdev:
Superseded by https://github.com/openmediavault/openmediavault/pull/909 because `60-persistent-storage.rules` will override the modified `ID_SERIAL` and `ID_SERIAL_SHORT` properties. Because of that `scsi_id` will export falys serial id and override the correct properties again. Because of that we need to go the old way.

PR #942, opened 2021-01-20
[LXC] Avoid "Device '/dev/dm-0' does not exist." problem

bestlong:
#944 

In Linux LXC /proc/partitions is the same as host, not match /dev

avoid like `Device '/dev/dm-0' does not exist.` problem.

Signed-off-by: Shao Yu-Lung <bestlong168@gmail.com>

- [ ] References issue

### DIFF
--- a/deb/openmediavault/usr/share/php/openmediavault/system/storage/backend/backendabstract.inc
+++ b/deb/openmediavault/usr/share/php/openmediavault/system/storage/backend/backendabstract.inc
@@ -134,7 +134,9 @@ abstract class BackendAbstract {
 		foreach(file("/proc/partitions") as $outputk => $outputv) {
 			if (1 !== preg_match($regex, $outputv, $matches))
 				continue;
-			$result[] = sprintf("/dev/%s", $matches[4]);
+			$deviceName = sprintf("/dev/%s", $matches[4]);
+			if (file_exists($deviceName))
+				$result[] = $deviceName;
 		}
 		return $result;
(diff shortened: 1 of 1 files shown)


deb/openmediavault/usr/share/php/openmediavault/system/storage/backend/backendabstract.inc
```diff
@@ -134,7 +134,9 @@ abstract class BackendAbstract {
 		foreach(file("/proc/partitions") as $outputk => $outputv) {
 			if (1 !== preg_match($regex, $outputv, $matches))
 				continue;
-			$result[] = sprintf("/dev/%s", $matches[4]);
+			$deviceName = sprintf("/dev/%s", $matches[4]);
+			if (file_exists($deviceName))
+				$result[] = $deviceName;
```
votdev, line 139: Could you please add a note here why this is done. It looks a little bit strange to check for a device that is listed in `/proc/partitions` without having the LXC context here.

bestlong: more info #944

votdev: Please add a note in the code, too.

votdev:
Thanks for the improvement. Superseded by https://github.com/openmediavault/openmediavault/pull/947.

PR #999, opened 2021-04-02
Refactor file system management.

votdev:
* /dev/disk/by-label filesystem device files are not supported anymore because they are not unique and predictable.
* Add 'comment' and 'usagewarnthreshold' to mountpoint datamodel.
* Try to auto-mount a filesystem if it is missing. Alert if the filesystem is still missing after a given time period.
* Adapt 'Storage | File Systems' pages.
* File system are not mounted immediately when they are added in the UI. They are mounted and can be used AFTER the configuration changes have been applied. This is the common workflow of the whole UI.

Filesystems itself can't be deleted anymore. If this is needed, the parent device can be wiped in the `Storage | Disks` page.

Bildschirmfoto vom 2021-04-03 13-51-00

Bildschirmfoto vom 2021-04-03 13-51-36

Bildschirmfoto vom 2021-04-03 13-53-14

Bildschirmfoto vom 2021-04-03 13-53-31

- [ ] References issue

### DIFF
--- a/deb/openmediavault/debian/changelog
+++ b/deb/openmediavault/debian/changelog
@@ -1,6 +1,10 @@
 openmediavault (6.0-1) UNRELEASED; urgency=low
 
   * Reimplement UI.
+  * /dev/disk/by-label filesystem device files are not supported
+    anymore because they are not unique and predictable.
+  * Change file system handling in UI.
+  * Adapt 'FileSystemMgmt' RPCs to new workflow.
 
  -- Volker Theile <volker.theile@openmediavault.org>  Mon, 14 Dec 2020 20:28:37 +0100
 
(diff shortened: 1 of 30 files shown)


votdev:
Is the `Delete` file system feature used by any plugin?

ryecoaaron:
> Is the Delete file system feature used by any plugin?

I don't think so.

ryecoaaron (dismissed):
Looks ok to me.  What happens if monit fails to mount a filesystem?

votdev:
> Looks ok to me. What happens if monit fails to mount a filesystem?

If that happens, then no automatic solution will work IMO. But if it happens, a notification is send in any case, so the administrator is aware of and can take over.

ryecoaaron:
> If that happens, then no automatic solution will work IMO. But if it happens, a notification is send in any case, so the administrator is aware of and can take over.

That works for me.

PR #1004, opened 2021-04-12
add code to skip adding debian-security on raspberrypi boards

ryecoaaron:
Skip adding debian security on Raspbian

Raspbian and RaspberryPi OS (not 64bit) rebuild packages from debian-security.  If they do not change the version and the debian-security repo is enabled, apt will have a package checksum mismatch.

Fixes: package checksum mismatches

Signed-off-by: Aaron Murray <plugins@omv-extras.org>
"""

The Signed-off-by line is important, and it is your certification that your contributions satisfy the developers certificate or origin.

For more examples, simply use "git log" and look at some historical commits.

This was just a quick overview. More information for contributors is available here:
https://docs.openmediavault.org/en/latest/development/contribute.html
-->

- [ ] References issue

### DIFF
--- a/deb/openmediavault/srv/salt/omv/deploy/apt/default.sls
+++ b/deb/openmediavault/srv/salt/omv/deploy/apt/default.sls
@@ -20,7 +20,7 @@
 {% set config = salt['omv_conf.get']('conf.system.apt.distribution') %}
 {% set use_kernel_backports = salt['pillar.get']('default:OMV_APT_USE_KERNEL_BACKPORTS', True) -%}
 {% set proxy_config = salt['omv_conf.get']('conf.system.network.proxy') %}
-{% set use_os_security = salt['pillar.get']('default:OMV_APT_USE_OS_SECURITY', False) %}
+{% set use_os_security = salt['pillar.get']('default:OMV_APT_USE_OS_SECURITY', True) %}
 
 {% set pkg_repos = [] %}
 {% for value in salt['pkg.list_repos']().values() %}
--- a/deb/openmediavault/srv/salt/omv/deploy/apt/default.sls
+++ b/deb/openmediavault/srv/salt/omv/deploy/apt/default.sls
@@ -20,7 +20,7 @@
 {% set config = salt['omv_conf.get']('conf.system.apt.distribution') %}
 {% set use_kernel_backports = salt['pillar.get']('default:OMV_APT_USE_KERNEL_BACKPORTS', True) -%}
 {% set proxy_config = salt['omv_conf.get']('conf.system.network.proxy') %}
-{% set rpi = salt['pillar.get']('default:OMV_IS_RPI', False) %}
+{% set use_os_security = salt['pillar.get']('default:OMV_APT_USE_OS_SECURITY', False) %}
 
 {% set pkg_repos = [] %}
 {% for value in salt['pkg.list_repos']().values() %}
--- a/deb/openmediavault/srv/salt/omv/deploy/apt/default.sls
+++ b/deb/openmediavault/srv/salt/omv/deploy/apt/default.sls
@@ -111,7 +111,7 @@ remove_apt_sources_list_security_{{ loop.index0 }}:
       - repo: "{{ security_pkg_repo.line }}"
 {% endfor %}
 
-{% if not rpi | to_bool %}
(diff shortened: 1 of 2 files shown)


deb/openmediavault/srv/salt/omv/deploy/apt/default.sls
```diff
@@ -20,6 +20,7 @@
 {% set config = salt['omv_conf.get']('conf.system.apt.distribution') %}
 {% set use_kernel_backports = salt['pillar.get']('default:OMV_APT_USE_KERNEL_BACKPORTS', True) -%}
 {% set proxy_config = salt['omv_conf.get']('conf.system.network.proxy') %}
+{% set rpi = salt['pillar.get']('default:OMV_IS_RPI', False) %}
```
votdev, line 23: Could we name this variable like this? Then it is equal to `OMV_APT_USE_KERNEL_BACKPORTS`.
```suggestion
{% set use_os_security = salt['pillar.get']('default:OMV_APT_USE_OS_SECURITY', False) %}
```

votdev (changes requested):

votdev:
Uahhh, i really hate such workarounds. To me it looks like RaspberryPi upstream is doing something terrible wrong.

ryecoaaron:
I understand but I don’t have a better solution.

deb/openmediavault/srv/salt/omv/deploy/apt/default.sls
```diff
@@ -20,6 +20,7 @@
 {% set config = salt['omv_conf.get']('conf.system.apt.distribution') %}
 {% set use_kernel_backports = salt['pillar.get']('default:OMV_APT_USE_KERNEL_BACKPORTS', True) -%}
 {% set proxy_config = salt['omv_conf.get']('conf.system.network.proxy') %}
+{% set use_os_security = salt['pillar.get']('default:OMV_APT_USE_OS_SECURITY', False) %}
```
votdev, line 23: Could you please set this to true? On a Raspberry installation this env variable must be set explicitly.

votdev (changes requested):

deb/openmediavault/srv/salt/omv/deploy/apt/default.sls
```diff
@@ -110,6 +111,8 @@ remove_apt_sources_list_security_{{ loop.index0 }}:
       - repo: "{{ security_pkg_repo.line }}"
 {% endfor %}
 
+{% if not use_os_security | to_bool %}
```
votdev, line 114:
```suggestion
{% if use_os_security | to_bool %}
```

deb/openmediavault/srv/salt/omv/deploy/apt/default.sls
votdev, line 23:
```suggestion
{% set use_os_security = salt['pillar.get']('default:OMV_APT_USE_OS_SECURITY', True) %}
```

votdev (changes requested):

votdev (approved):

votdev:
@ryecoaaron I stumbled over this code [here](https://github.com/openmediavault/openmediavault/blob/5.x/deb/openmediavault/srv/salt/omv/deploy/apt/default.sls#L105) and wonder why your PR is necessary. Salt should remove the `debian-security` repositories spread across other APT config files, so only the `/etc/apt/sources.list.d/openmediavault-debian-security.list` should exist afterwards.

rabelux:
This feature has been released, right?
Is there already an implementation of setting the env-var?

I ask because on my raspberry pi (with OMV 5.6.7) the security-repo got readded yesterday. I had removed it about a month ago.

`omv-env list` shows that the variable `OMV_APT_USE_OS_SECURITY` is known, but has no value as
`omv-env get OMV_APT_USE_OS_SECURITY` returns nothing.

I guess l can set this variable to False with `omv-env set OMV_APT_USE_OS_SECURITY False` ?

ryecoaaron:
Yes, it has been released.  The only thing that sets it is the install script.  If you have an existing install, you will need to run the omv-env command to set it.  omv-env set OMV_APT_USE_OS_SECURITY False

rabelux:
Did that, thanks for your response 👍

PR #1005, opened 2021-04-13
Do not use {{ _selected[0].xyz }} for datatable actions

votdev:
... that support multi-selection. Use `{{ xyz }}` instead. `{{ _selected[0].xyz }}` can still be used for actions that support only single-selection. The intention of using `_selected` was to show plugin devs how to process the values of the selected row(s).

- [ ] References issue


ryecoaaron:
Was using the {{ xyz }} syntax already - https://github.com/OpenMediaVault-Plugin-Developers/openmediavault-developer/blob/master/usr/share/openmediavault/workbench/component.d/omv-services-developer-plugins-datatable-page.yaml#L68.  So, that is fine with me.

Issue #1040, opened 2021-06-07
Replace custom php_pam with now supported version from PECL


mi-hol:
**Is your feature request related to a problem? Please describe.**
In the past php_pam was no longer supported therefore OMV added a custom version in commit https://github.com/openmediavault/openmediavault/commit/ce9c3409b56102bbc88b99223971607138d0ed7c
But this custom version creates a maintenance burden and PHP version dependency

**Describe the solution you'd like**
Replace custom php_pam with now supported version from [PECL](https://pecl.php.net/package/PAM)

**Additional context**
new php_pam maintainer assigned in PECL https://github.com/amishmm/php-pam/issues/3

PR #1041, opened 2021-06-09
Replace custom php_pam with now supported version from PECL

votdev:
Fixes: https://github.com/openmediavault/openmediavault/issues/1040

- [x] References issue


mi-hol:
Wow! that was a fast fix.
Are there other modules that I could help with?

votdev:
> Are there other modules that I could help with?

No, thx. That was the only one that has the be maintained by OMV.

PR #1061, opened 2021-07-01
Add option to show the path of the shared folder or mount point with which this field is associated

votdev:
Peek 2021-07-10 18-48

- [ ] References issue

### DIFF
--- a/deb/openmediavault/workbench/src/app/core/components/limn-ui/form/components/form-folderbrowser/form-folderbrowser.component.html
+++ b/deb/openmediavault/workbench/src/app/core/components/limn-ui/form/components/form-folderbrowser/form-folderbrowser.component.html
@@ -1,3 +1,16 @@
+<mat-form-field *ngIf="config.dirVisible"
+                fxFlex="25"
+                [formGroup]="formGroup">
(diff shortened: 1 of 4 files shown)


ryecoaaron (dismissed):
As long as I an get the base path into php, that is fine with me.

votdev:
> As long as I an get the base path into php, that is fine with me.

The form field will not work that way, but you can modify the value in the post request configuration:

```
request:
      service: Developer
      get:
        method: getSettings
      post:
        method: setSettings
        params:
          reldirpath: '/{{ reldirpath }}'
```

ryecoaaron:
> Relative paths never start with a / according to https://www.geeksforgeeks.org/absolute-relative-pathnames-unix/. Breaking this rule here will break other things that rely on this behaviour and will finally result in paths like /src/xxxx//foo/var/ (which isn't wrong at the end, but is not correct).

I know relative paths shouldn't start with /.  I was just wanted a better visual method of picking a root path.  If they picked /srv/path and the field shows srv/path, I think that is confusing.  Having the base path makes it a little better.  There are just some services/plugins that don't work well with shared folders.

PR #1135, opened 2021-11-12
Update etc-exports.j2

hypnotoad:
use share's uuid as fsid. This leads to consistend fsid entries across restarts and configuration changes of other shares. Clients will now not have any stale file handles any more.

https://forum.openmediavault.org/index.php?thread/40749-fsid-seems-seems-to-be-applied-dynamically/&postID=286858#post286858

Signed-off-by: Ralf Dragon <rd-comm@lindra.de>

thanks to votdev for the actual code!

- [ ] References issue


votdev:
Thx

PR #1174, opened 2022-01-01
Add binary unit form field.

votdev:
Example:
```
{
            type: 'binaryUnitInput',
            name: 'recyclemaxsize',
            label: gettext('Maximum file size'),
            hint: gettext(
              'Files that are larger than the specified number will not be put into the recycle bin.'
            ),
            value: 0,
            modifiers: [
              {
                type: 'disabled',
                constraint: { operator: 'falsy', arg0: { prop: 'recyclebin' } }
              }
            ],
            validators: {
              min: 1073741824,
              max: 2147483648,
              required: true
            },
            flex: 45
          }
```

Peek 2022-01-01 20-13

- [ ] References issue


votdev:
@ryecoaaron Just for your interest. Maybe you need such an input field somewhere in a plugin.

PR #1175, opened 2022-01-03
Prevent the installation of OMV in a desktop environment.

votdev:
This PR will help to reduce maintenance in the forum and should prevent users from having a bad experience with OMV.

- [ ] References issue

### DIFF
--- a/deb/openmediavault/debian/control
+++ b/deb/openmediavault/debian/control
@@ -8,6 +8,7 @@ Homepage: http://www.openmediavault.org
 
 Package: openmediavault
 Architecture: all
+Conflicts: gdm3, sddm, lxdm, xdm, lightdm, slim, wdm
 Depends: php-fpm, libpam-modules, php-json, dpkg, patch,
  php-cgi, php-cli, php-mbstring, php-pam, sudo, ethtool, python3-dialog,
  acl, iproute2, xfsprogs, jfsutils, ntfs-3g, hdparm, wsdd (>= 0.6.2),


votdev:
@ryecoaaron

mi-hol:
Will the change "Conflicts: gdm3, sddm, lxdm, xdm, lightdm, slim, wdm" work for SBCs too?
If yes, the separate implementation in install script could & should be removed

votdev:
> Will the change "Conflicts: gdm3, sddm, lxdm, xdm, lightdm, slim, wdm" work for SBCs too? 

It should work everywhere. 

> If yes, the separate implementation in install script could & should be removed

The separate check in the install script should be retained because it has a more human readable error message than the error that will be thrown by APT. It is also a better user experience if the installation aborts immediately than in the middle of the installation (with a cryptic message).

votdev:
The installation will abort like this:
```
dpkg -i openmediavault_6.0.7-1_all.deb 
dpkg: regarding openmediavault_6.0.7-1_all.deb containing openmediavault:
 openmediavault conflicts with xdm
  xdm (version 1:1.1.11-3+b2) is present and installed.

dpkg: error processing archive openmediavault_6.0.7-1_all.deb (--install):
 conflicting packages - not installing openmediavault
Errors were encountered while processing:
 openmediavault_6.0.7-1_all.deb
```

ryecoaaron:
I will leave it in the script because it will stop people from blindly saying yes.

mi-hol:
> The separate check in the install script should be retained because it has a **more human readable error message**

ok, makes sense.
Now 2 technical implementation that need to be in sync in 2 different repositories usually ask for trouble in the long run.
Could we find a better approach that eliminates this risk?

votdev:
> > The separate check in the install script should be retained because it has a **more human readable error message**
> 
> 
> 
> ok, makes sense.
> 
> Now from experience, 2 technical implementation that need to be in sync in 2 different repositories usually asks for trouble in the long run.

Surely not.

> 
> Could we find a better approach that eliminates this risk?

No further action is needed.

ryecoaaron:
New window managers don't come along very often.  If one does, then the person will be able to install it.  Not any worse than it was before.  If one of them goes away, it doesn't cause any problems.  No need to make this a bigger issue than it is.

darrepac:
> The installation will abort like this:
> 
> ```
> dpkg -i openmediavault_6.0.7-1_all.deb 
> dpkg: regarding openmediavault_6.0.7-1_all.deb containing openmediavault:
>  openmediavault conflicts with xdm
>   xdm (version 1:1.1.11-3+b2) is present and installed.
> 
> dpkg: error processing archive openmediavault_6.0.7-1_all.deb (--install):
>  conflicting packages - not installing openmediavault
> Errors were encountered while processing:
>  openmediavault_6.0.7-1_all.deb
> ```

I have omv5 installed with debian desktop environment. Why? I don't remember but I don't need it. I just want a path to update to OMV6 by removing desktop environment.
Will the upgrade script (omv-release-upgrade) will abort or it will propose to remove conflicted packages?
If not removing, do I have to remove manually before to launch upgrade script?

ryecoaaron:
Yes remove xdm before running omv-release-upgrade

WarlockFish:
if i installed openmediavault , then i need a simple desktop environment .this desktop environment package is Conflicts.because this control's `Conflicts:` key.  

i try chang this package control ，and build package ,  install openmediavault , then  install lightdm .  the openmediavault is running  normally。 

```
Conflicts: gdm3, sddm, lxdm, xdm, lightdm, slim, wdm
``` 

i does not like this way ， it need change package 。

i installed  openmediavault  ，then also install desktop environment ,    what shoule i do ?

PR #1187, opened 2022-01-19
Improve watchdog SLS

votdev:
- Ensure the kernel module is unloaded if watchdog gets disabled.
- Switch to default timeout of 15 seconds because of problems (endless boot loop) on RPi boards.

- [ ] References issue

### DIFF
--- a/deb/openmediavault/srv/salt/_grains/omv.py
+++ b/deb/openmediavault/srv/salt/_grains/omv.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+#
 # This file is part of OpenMediaVault.
 #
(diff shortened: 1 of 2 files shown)


votdev:
@ryecoaaron You can test it the following way.
```
$ omv-salt stage run prepare
$ salt-call grains.get is_rpi
```

ryecoaaron:
Seems to be working well.  Installing the 6.0.10 build didn't cause the RPi to boot loop.  Grain looks good too.

aaron@rpi4k:~ $ sudo salt-call grains.get is_rpi
local:
    True

Thanks!

ryecoaaron (approved):
Looks good.

PR #1183, opened 2022-01-13
Use systemd watchdog instead of separate daemon

votdev:
Fixes: https://github.com/openmediavault/openmediavault/issues/224
Fixes: https://github.com/openmediavault/openmediavault/issues/1167

Signed-off-by: Volker Theile <votdev@gmx.de>

- [x] References issue
- [ ] Includes tests for new functionality or reproducer for bug

PR #1188, opened 2022-01-20
Need to set the watchdog timeout to <= 15s for all ARM devices

votdev:
Fixes: https://github.com/openmediavault/openmediavault/pull/1183

- [x] References issue


ryecoaaron (changes requested):
I think the osarch grain should be used instead of parsing the cpuarch grain.  armhf and arm64 could use the lower 15 second time.

ryecoaaron:
Looks good now.

PR #1195, opened 2022-01-24
Relocate the user management navigation menu

votdev:
navbar

- [ ] References issue

### DIFF
--- a/deb/openmediavault/debian/changelog
+++ b/deb/openmediavault/debian/changelog
@@ -2,6 +2,7 @@ openmediavault (6.0.9-4) unstable; urgency=low
 
   * Update locale files.
   * Use systemd watchdog instead of separate daemon.
+  * Relocate the user management navigation menu.
   * Issue #1181: Use 'Warning' color for SMART assessment status
     'BAD_SECTOR'.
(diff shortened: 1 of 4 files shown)


votdev:
Which position of the `User Management` menu is the best from the end-user point of view? Which position will help to minimize problems for them.

Please start a discussion here.

PrplHaz4:
Makes sense at the bottom for me since I consider it the last step of setting things up.

ryecoaaron:
Seems fine to me.  What about changing it to just "Users" to make the side bar narrower?  Management is implied in most of the other sections.

flmaxey:
As previously note in a PM:
If **User Management** is in the middle of the absolute bare minimum of configuration and services required to setup a working NAS (with open permissions), it becomes a temptation to new users.  (What does this do...?)  I like the idea of getting new users started without incident.  

I believe moving **User Management** down the list would have at least a small damping effect on new users opening it to see what's there and "clicking on boxes".  Once OMV is up and running, I believe the average user would be reluctant to tamper with a working NAS or, at least, they'll understand the effects of indiscriminate tampering.

Where advanced and expert users are concerned, it really doesn't matter where the menu item is located.  They'll recognize it and use it as needed.

PR #1199, opened 2022-01-30
Disable autocapitalize

m0n4:
Disable autocapitalize  for the user field on the login page.

Allows to avoid that the virtual keyboards put a capital letter to the first word by default. 
This doesn't prevent you from changing the case of the virtual keyboard.
[https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize)

Fixes: None

Signed-off-by: Trivial patch, CLA not needed

- [ ] References issue

### DIFF
--- a/deb/openmediavault/workbench/src/app/core/pages/login-page/login-page.component.ts
+++ b/deb/openmediavault/workbench/src/app/core/pages/login-page/login-page.component.ts
@@ -53,6 +53,7 @@ export class LoginPageComponent implements OnInit {
         label: gettext('User name'),
         autofocus: true,
         autocomplete: 'username',
+        autocapitalize: 'none',
         icon: Icon.user,
         validators: {
           required: true


votdev:
I'll close this PR because of:

- A CLA is always needed
- The code is incorrect and does not work

I've created a new PR based on the hint you gave with your PR. See https://github.com/openmediavault/openmediavault/commit/260c5f255a312b693bffc975f24d7210a5917ff3 for more information.

m0n4:
No problems, thanks for fixing it.
I thought it was trivial in the sense of documentation: [Trivial patches like spelling fixes or missing words won’t require an agreement](https://openmediavault.readthedocs.io/en/5.x/development/contribute.html#can-i-submit-patches-without-having-signed-the-cla)

PR #1226, opened 2022-02-23
change Link to Blog (to avoid loading subdomain with invalid SSL-cert)

nextgen-networks:
There seems to be an old(?) invalid subdomain mentioned in README.md.
https://blog.openmediavault.org leads to an SSL error because *.1blu.de wildcard cert is used.
Instead please use the link to blog that is already mentioned on OMV-Website navigation: https://www.openmediavault.org/blog.html

And what also needs to be told:
BIIIIG THANKS to Volker and all the contributors to this fantastic project!

- [ ] References issue


votdev:
The issue has been fixed, so the subdomain works again.

PR #1264, opened 2022-03-28
Add SSL support via reverse proxy (Traefik)

votdev:
To keep the Salt code simple, the reverse proxy is also used in HTTP mode.

- [ ] References issue


PrplHaz4:
A quick look at the code makes it seem specific to photoprism. I’d expect a reverse proxy to be utilized by all services on a system, including letsencrypt certs in front of the omv web app. 

I recognize making it general purpose is no small feat so starting with a single service makes sense. 

In docker traefik can use labels for auto configuration but not sure about podman….

votdev:
> A quick look at the code makes it seem specific to photoprism. I’d expect a reverse proxy to be utilized by all services on a system, including letsencrypt certs in front of the omv web app.
> 
> I recognize making it general purpose is no small feat so starting with a single service makes sense.
> 
> In docker traefik can use labels for auto configuration but not sure about podman….

Adding the reverse proxy as a separate container in the pod is by intention. If the user selects SSL encryption, the pod will NOT expose HTTP, but only HTTPS. This is a security decision.

All official OMV plugins that are using containers/podman will use this setup in future.

votdev:
> > A quick look at the code makes it seem specific to photoprism. I’d expect a reverse proxy to be utilized by all services on a system, including letsencrypt certs in front of the omv web app.
> > I recognize making it general purpose is no small feat so starting with a single service makes sense.
> > In docker traefik can use labels for auto configuration but not sure about podman….
> 
> Adding the reverse proxy as a separate container in the pod is by intention. If the user selects SSL encryption, the pod will NOT expose HTTP, but only HTTPS. Using a container in the pod ensures that only the dedicated port and protocol are accessible to the host. This is a security decision.
> 
> All official OMV plugins that are using containers/podman will use this setup in future.

PR #1268, opened 2022-04-02
Add hint about port 465

Jip-Hop:
Add hint for the third common SMTP port.

Signed-off-by: Jip de Beer 2871973+Jip-Hop@users.noreply.github.com


votdev:
Thanks for all the effort, but translations work a little bit different in OMV. You need to modify the string in the UI code, a special command extracts the translation messages and build the *.pot file which is transferred to Transifex where the translators need to translate it. Afterwards the translations are pulled from Transifex and stored in the GitHub repo.

votdev:
I've closed the PR because i've done the changes the right way. It's not possible to upload translations without to Transifex without the risk to loose already translated strings.

PR #1279, opened 2022-04-16
Fix getRootDeviceFile not returning the correct parent device when OS is installed on a  full disk encrypted drive

myomv100:
Hello Everyone,

This pull request goal is to fix the function getRootDeviceFile not returning the correct parent device when the OS is installed on a full disk encrypted drive.
The previous implmenetation returns the partition name of the root directory not the device a the function name implies.:
```
public static function getRootDeviceFile() {
...
...
return $mp->getDeviceFile();  -> this just returns the parition of the mount point not the device file
```

it's optimal in a case where the OS is installed on top of a normal parition not an LVM :
image

Now when you install OMV on a full disk encrypted disk, debian installer creates an LVM group on top of a partiton and then makes differnet volumes to serve different purposes ( swap, root, ..) :
image

What I did is to navigate through the tree up to the parent device not just a parition of the mount point.

I intrduced a new function getPartitionParentDevice and modified the getRootDeviceFile function to return a device not a partiton.

Thank you

### DIFF
--- a/deb/openmediavault/debian/changelog
+++ b/deb/openmediavault/debian/changelog
@@ -1,6 +1,6 @@
 openmediavault (5.6.27-1) stable; urgency=low
 
-  * 
+  * Quote names in quota SLS to support blanks in user/group names.
 
  -- Volker Theile <volker.theile@openmediavault.org>  Tue, 01 Mar 2022 19:53:05 +0100
 
--- a/deb/openmediavault/srv/salt/omv/deploy/quota/default.sls
+++ b/deb/openmediavault/srv/salt/omv/deploy/quota/default.sls
@@ -91,13 +91,13 @@ quota_on_{{ fsuuid }}:
 {% for usrquota in quota.usrquota %}
 quota_set_user_{{ fsuuid }}_{{ usrquota.name }}:
   cmd.run:
-    - name: setquota --user {{ usrquota.name }} {{ usrquota.bsoftlimit }} {{ usrquota.bhardlimit }} {{ usrquota.isoftlimit }} {{ usrquota.ihardlimit }} {{ device }}
+    - name: setquota --user '{{ usrquota.name }}' {{ usrquota.bsoftlimit }} {{ usrquota.bhardlimit }} {{ usrquota.isoftlimit }} {{ usrquota.ihardlimit }} {{ device }}
 {% endfor %}
 
 {% for grpquota in quota.grpquota %}
 quota_set_group_{{ fsuuid }}_{{ grpquota.name }}:
   cmd.run:
-    - name: setquota --group {{ grpquota.name }} {{ grpquota.bsoftlimit }} {{ grpquota.bhardlimit }} {{ grpquota.isoftlimit }} {{ grpquota.ihardlimit }} {{ device }}
+    - name: setquota --group '{{ grpquota.name }}' {{ grpquota.bsoftlimit }} {{ grpquota.bhardlimit }} {{ grpquota.isoftlimit }} {{ grpquota.ihardlimit }} {{ device }}
 {% endfor %}
 
 {% endif %}


votdev:
Thanks for your contribution. Is there a way to retrieve the necessary way in a more cheaper way? The problem is that calling an executable is a really expensive action. The function you've modified is called very often, so it may increase the execution time of a RPC request in a dramatic way if there are more than one or two disk devices installed in the system. Keep in mind that there are users with > 25 disks and more.

myomv100:
it actually does increase the execution time. It runs once for every drive attached to the system.  We could execute the shell command once and pass the results to isRootDeviceFile

isRootDeviceFile is used in different location. There might be some way to improve the \OMV\System\Storage\StorageDevice::enumerate function so we pass the output of it to isRootDeviceFile

myomv100:
it actually does increase the execution time. It runs once for every drive attached to the system.  We could execute the shell command once and pass the results to isRootDeviceFile

isRootDeviceFile is used in different location. There might be some way to improve the \OMV\System\Storage\StorageDevice::enumerate function so we pass the output of it to isRootDeviceFile

myomv100:
Replacing this line :
`"isroot" => \OMV\System\System::isRootDeviceFile($sd->getCanonicalDeviceFile(), FALSE)
`
with : 
`"isroot" => $rootDeviceFile === basename($sd->getCanonicalDeviceFile()) ? TRUE : FALSE
`
$rootDeviceFile  is already defined at the begning of enumerateDevices in diskmgmt :
`$rootDeviceFile = \OMV\System\System::getRootDeviceFile();
`

This seems to solve the issue of calling the executable multiple times. We moved the comparison from the storage backend to be done in the rpc APIs. So no need for isRootDeviceFile function at all. All we need is getRootDeviceFile. The later function tells us only what the root device is. There is no need to pass anything and check everytime with getRootDeviceFile.

deb/openmediavault/usr/share/php/openmediavault/system/system.inc
```diff
@@ -28,6 +28,31 @@ require_once("openmediavault/functions.inc");
...
  */
 class System {
+	/**
+	 * Get a partition parent block device
+	 * @return Returns the storage parent device file of the partiton of where the
+	 * operating system is installed
+	 * @throw \OMV\ExecException
+	 */
+	public static function getBlockParentDevice($blockdevices, $partiton) {
+		static $blockParentDevice = NULL;
```
votdev, line 38: Why is this a static variable?

deb/openmediavault/usr/share/openmediavault/engined/rpc/diskmgmt.inc
```diff
@@ -90,8 +90,9 @@ class DiskMgmt extends \OMV\Rpc\ServiceAbstract {
 				"serialnumber" => $sd->getSerialNumber(),
 				"wwn" => $sd->getWorldWideName(),
 				"israid" => $sd->isRaid(),
-				"isroot" => \OMV\System\System::isRootDeviceFile(
-					$sd->getCanonicalDeviceFile(), FALSE)
+				"isroot" =>
+				$rootDeviceFile === basename($sd->getCanonicalDeviceFile()) ?
```
votdev, line 94: Why using basename here?

deb/openmediavault/usr/share/php/openmediavault/system/system.inc
```diff
@@ -66,11 +98,13 @@ class System {
 			return TRUE;
 		$rootDeviceFile = self::getRootDeviceFile();
 		if (TRUE === $exact) {
-			return (($rootDeviceFile == $deviceFile) ||
+			return ((basename($rootDeviceFile) == basename($deviceFile)) ||
```
votdev, line 101: Why using `basename` here? How does the `$deviceFile` may look like in your tests? I'm asking that to understand what you are trying to approach here.

votdev: Does `basename` will not counteract the intention of `$exact === TRUE` here?

votdev (changes requested):
Please post the whole output of `lsblk -J -o NAME` of your test system. I will understand what you are trying to achieve here. Please also post the output of `mount`. 
I'm also wondering why you are using `basename($mp->getDeviceFile())`. How does the root device file look like? Do you expect a device file like `/dev/foo/bar` here?

Right at the moment i do not have a good feeling with this PR. I assume it will break existing systems.

votdev:
Please create this PR against `master`. If there is a need to backport it to older versions, this will be done in a separate backport PR. Note, no new features will be implemented in OMV5; and your PR is a feature. So expect it to be available in OMV6 only.

deb/openmediavault/usr/share/php/openmediavault/system/system.inc
myomv100: Out put of` lsblk -J -o NAME` command
```
root@openmediavault:~# lsblk -J -o NAME
{
   "blockdevices": [
      {"name":"sda",
         "children": [
            {"name":"sda1"},
            {"name":"sda2"},
            {"name":"sda5",
               "children": [
                  {"name":"sda5_crypt",
                     "children": [
                        {"name":"nas--vg-root"},  <---- Here is my root mount point /
                        {"name":"nas--vg-swap_1"}
                     ]
                  }
               ]
            }
         ]
      },
      {"name":"sdb"},
      {"name":"sdc"},
      {"name":"sr0"}
   ]
}
```

I used `basename($mp->getDeviceFile())` to compare only block devices names and exclude the full path. 

The goal was to tell definitley the root device no matter how deep it's buried inside an lvm group or no matter what its name was as long as it's associated with / mount point it should be considered a root device.  

I thought of adding a label to root device instead. And look for it in `getRootDeviceFile`. I did not prefer adding a label because there was no way to prevent other devices from being labeled with the same label name.

myomv100: `static $blockParentDevice = NULL` 
 Prevents $blockParentDevice variable state from being overwritten or reset to NULL by instantiation due to recursion.
  
`$blockParentDevice = $items[0];`   <--- this should happens only once. It's not possible to have different devices to be mounted at mount point  /

deb/openmediavault/usr/share/openmediavault/engined/rpc/diskmgmt.inc
myomv100: Because `getBlockParentDevice ` returns only device name to `getRootDeviceFile `, therefore `getRootDeviceFile` will also returns device name without path. We compare only names.

deb/openmediavault/usr/share/php/openmediavault/system/system.inc
myomv100: `isRootDeviceFile ` function may not be used anymore at all. 

Agree, that added line is not necessary so does the function itself.

myomv100:
Just found out that USBBackup plugin is using `isRootDeviceFile `
This requires redoing things without breaking how OMV is working. .. working on it

PR #1291, opened 2022-04-25
Fix getRootDeviceFile not returning the correct parent device when the OS is installed on an LVM or full disk encrypted drive

myomv100:
Hello Everyone,

This pull request goal is to fix the function getRootDeviceFile not returning the correct parent device when the OS is installed on a full disk encrypted drive.
The previous implmenetation returns the partition name of the root directory not the device a the function name implies.:
```
public static function getRootDeviceFile() {
...
...
return $mp->getDeviceFile();  -> this just returns the parition of the mount point not the device file
```

it's optimal in a case where the OS is installed on top of a normal parition not an LVM :
image

Now when you install OMV on a full disk encrypted disk, debian installer creates an LVM group on top of a partiton and then makes differnet volumes to serve different purposes ( swap, root, ..) :
image

What I did is to navigate through the tree up to the parent device not just a parition of the mount point.

I intrduced a new function getPartitionParentDevice and modified the getRootDeviceFile function to return a device not a partiton.

Thank you

**This is a PR against master with less modification on OMV.
Prevous PR was closed due to reverting back  changes on 5.x  and adding them master.
https://github.com/openmediavault/openmediavault/pull/1279**

### DIFF
--- a/deb/openmediavault/usr/share/php/openmediavault/system/system.inc
+++ b/deb/openmediavault/usr/share/php/openmediavault/system/system.inc
@@ -29,6 +29,27 @@ require_once("openmediavault/functions.inc");
  */
 class System {
 	/**
+	 * Get a partition parent block device
+	 * @return Returns the storage parent device file of the partiton of where the
+	 * operating system is installed
+	 * @throw \OMV\ExecException
+	 */
+	public static function getBlockParentDevice($blockdevices, $partiton, &$rootDeviceFile) {
+		foreach ($blockdevices as $blocks => &$block) {
+			foreach ($block->children as $children => &$child) {
+				$child->name = $block->name . ":" . $child->name ;
+			}
+			$items = explode(":",$block->name);
+			if (strpos($partiton, end($items)) === 0) {
+				$rootDeviceFile = $items[0];
+			} else if(!is_null($block->children)) {
+				self::getBlockParentDevice(
+					$block->children, $partiton, $rootDeviceFile);
+			}
+		}
+	}
+
+	/**
 	 * Get the device file where the operating system is installed on (e.g.
 	 * /dev/sda1). To avoid further useless process calls to determine the
 	 * root device file the value is cached in a static variable when this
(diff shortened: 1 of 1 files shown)


votdev:
Could you please add some PHP unit tests for those modified and new functions. I think this is necessary to do not introduce a regression and to prevent fatal consequences on production systems out there. When implementing these unit tests, please do not focus on your new functionality, the old behaviour need to be tested as well to ensure this is not broken.

mi-hol:
Adding a unit test makes perfect sense.
Would the file deb/openmediavault/usr/share/openmediavault/unittests/php/test_openmediavault_functions.php be the correct place to add it?

votdev:
> Adding a unit test makes perfect sense. Would the file deb/openmediavault/usr/share/openmediavault/unittests/php/test_openmediavault_functions.php be the correct place to add it?

No, https://github.com/openmediavault/openmediavault/blob/master/deb/openmediavault/usr/share/openmediavault/unittests/php/test_openmediavault_system_system.php is the better place for that.

myomv100:
[https://github.com/openmediavault/openmediavault/blob/master/deb/openmediavault/usr/share/openmediavault/unittests/php/test_openmediavault_system_system.php
](url)
From the added unit tests. we can see that you are testing if:
1. root mount point resides on the block without partition
2. root mount point is on a partition

Now we need to unit test the function of which the system is installed on top of an lvm group which is expected when debian has luks encryption enabled for the root file system see:
image

myomv100:
https://github.com/openmediavault/openmediavault/blob/master/deb/openmediavault/usr/share/openmediavault/unittests/php/test_openmediavault_system_system.php 

Test getRootDeviceFile does not override the static function. Maybe because getRootDeviceFile calls another static function. Not sure.

mi-hol:
ping @votdev

mi-hol:
@votdev does closing mean the issue is solved by a change you implemented?

votdev:
No, it's been closed because of inactivity. Such a feature that touches an important part of the core engine requires a whole bunch of unit tests to ensure that it will not break anything. This was not the case here.

mi-hol:
> closed because of inactivity

Well there was an unanswered question. Is it good style to not respond and then state inactivity as a reason?

votdev:
I deleted the previous post because I felt attacked as I make my apologies.

votdev:
Please open a new pull request including a bunch of unit tests (by mocking real scenarios). The possibility that this feature breaks existing setups is really high. If a pull request does not ensure the functionality of such a critical part it will not be merged.

Issue #1326, opened 2022-06-14
yaml highlighting text area component


ryecoaaron:
**Is your feature request related to a problem? Please describe.**
It would be nice to have a yaml-highlighting text area for plugins to use.

**Describe the solution you'd like**
A text area component that could be added to a form or datatable page that would highlight yaml code for editing.

**Describe alternatives you've considered**
Using normal text area now.

votdev:
After some research these are the problems:
- The Monaco editor is not supported in mobile browsers or mobile web frameworks. 
- A <textarea> form field does not support customization of the displayed content.

This means if we want to support syntax highlighting as an editor (and not only to display it) it is necessary to implement a new Angular form control which is using [HighlightJS](https://highlightjs.org/) or [PrismJS](https://prismjs.com/) (the later is my favorite).

votdev:
Good examples for PrismJS:
- https://techincent.com/code-syntax-highlighter-angular-with-prism-js/
- https://github.com/angular-patterns/ng-prismjs

votdev:
Looking what other projects are using (e.g. Portainer), i found [CodeMirror](https://codemirror.net) for which an [Angular component](https://github.com/scttcper/ngx-codemirror) exists.

PR #1327, opened 2022-06-18
Add 'codeEditor' form field

votdev:
Add code editor with syntax highlighting.

CodeMirror 6:
Peek 2022-06-20 00-13

Monaco (outdated):
Peek 2022-06-18 21-30

Fixes: https://github.com/openmediavault/openmediavault/issues/1326

- [x] References issue


votdev:
I will switch from Monaco to CodeMirror6 incl. writing my own MatFormFieldControl because i do not see any progress in migrating the ngx-monaco-editor component to Angular 13. If it is not migrated, this is a blocker when migrating OMV to Angular >= 13. Monaco looks great, but i think CodeMirror is more lightweight.

ryecoaaron:
This is great.  I can use it for a few plugins.  Looks very good but when I type, I am getting mirrored text moving from right to left.
image

PR #1338, opened 2022-07-01
Replace custom cursors with system provided alternatives

SkaceKamen:
Replaces custom cursor images with cursors provided by the browser. The custom icons don't scale well for non-standard DPI. Also the waiting icon has incorrect aspect-ratio and is very pixelated. And I think using OS provided cursors is better user experience.

Removed icons:
image
image

Replaced with (rendered on Windows 10):
image
image


ryecoaaron:
The cursors are like that on purpose.  They are a tribute to the Amiga.  Pretty sure Volker does not want to change them.

votdev:
The cursors will not be changed because this is a tribute to the Amiga computer.

SkaceKamen:
Ok, sorry, I thought it was leftover from something. Just a question, is the 503 error screen also a tribute? It seems a bit out of a place design-wise.

votdev:
> Ok, sorry, I thought it was leftover from something. Just a question, is the 503 error screen also a tribute? It seems a bit out of a place design-wise.

Yes, see https://en.wikipedia.org/wiki/Guru_Meditation

PR #1406, opened 2022-10-16
 Add iOS webapp support

neheb:
This way when pinned to Springboard, Openmediavault opens in its own
instance instead of a Safari tab.


votdev:
See https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html

neheb:
Not sure what you mean.  It seems to work fine.

votdev:
> Not sure what you mean. It seems to work fine.

I have added the link only for reference.

PR #1454, opened 2022-12-29
Add a tag input form field.

votdev:
The form value (which is later stored in the database) will look like `foo,bar`.

Peek 2022-12-29 15-46

- [ ] References issue


votdev:
@ryecoaaron Maybe the tag support is also useful for one or the other plugin.

Plan is to re-use the existing `comments` field on various pages.

ryecoaaron (approved):
Seems like it would useful on a some plugins.

Issue #1310, opened 2022-05-24
Dashboard settings should be persisted in file or database instead of cookies


TheMelmacian:
**Is your feature request related to a problem? Please describe.**
Currently the dashboard configuration is stored in cookies (see https://forum.openmediavault.org/index.php?thread%2F40847-dashboard-doesn-t-remember-settings%2F=).
This can result in different settings on different devices (PC, Laptop, phone etc.) or loss of the settings if the cookies are deleted.

**Describe the solution you'd like**
All settings should be persisted in a file or database so it is synced between devices and different browsers and is not lost when the cookies are deleted.

godfuture:
Same here. I often use private browsing for locally hosted services in LAN to prevent browsing history being filled up. Now with OMV6 I end up with empty dashboard quite often. As much development in browser tech was focused on cookie management and cockie blocking, I think we should not store these settings in cookies.

Why not storing the dashboard in user settings and OMV is identifying the browser agent to separate mobile from desktop enviroment. This is done in a lot of website already. If someone wants to have even more ways to define different dashboards, he could add another admin user. 

But I guess most of the users are served if we have desktop and mobile dashboards within the same user.

fire-burner:
Privacy oriented users often have their cookies deleted when closing their bowsers. Dashboard settings should not be stored in cookies, but in a config file instead. It is rather annoying to always have to reconfigure my dashboard when I now and then take a look at my OMV machine.

ryecoaaron:
You can set exceptions for cookie deletion.  And feel free to submit a PR with the change to client side settings stored in a file on the server.  This change should allow different settings for different clients since I don't want the same widgets for every client - phone vs laptop.

fire-burner:
That is fair point too. But I am not sure, how many are using different devices typed to regularly access the web frontend.

ryecoaaron:
Many are using both phone and computer often if comments in the forum mean anything.

DjoeC:
I'm in support of storing dashboard settings serverside instead of clientside.  All my sites run as cookieless as possible including autodelete 15 seconds after the tab has been closed since it's not possible to crossprotect between windows.  I'd rather change dashboard settings when needed than having to reconfigure the dashboard every time I login.

lessload:
Such a stupid logic to keep user setting in cookies.!! That why many people move to TrueNAS.

votdev:
[...]

PR #1463, opened 2023-01-09
Dashboard settings should be persisted in file or database instead of cookies

jpspringall:
Fixes: https://github.com/openmediavault/openmediavault/issues/1310

- [x] References issue


deb/openmediavault/debian/openmediavault.dirs
```diff
@@ -6,6 +6,7 @@ usr/share/php/openmediavault
 var/lib/openmediavault
 var/lib/openmediavault/apt
 var/lib/openmediavault/cron.d
+var/lib/openmediavault/dashboard.d
```
jpspringall, line 9: I am presuming that adding this means that the directory will get automatically created?

For my local dev, i manually created the folder

.gitignore
```diff
@@ -1 +1,2 @@
 .idea/
+.vscode/
```
jpspringall, line 2: Added this as i use vscode

deb/openmediavault/workbench/proxy.conf.json
```diff
@@ -13,5 +13,10 @@
     "target": "http://omv6box.local/",
     "secure": false,
     "logLevel": "debug"
+  },
```
jpspringall, line 16: I added this, so that it can get the navigation info etc from vagrant when running locally

deb/openmediavault/workbench/src/app/core/services/dashboard-widget-config.service.ts
```diff
@@ -76,23 +85,74 @@ export class DashboardWidgetConfigService {
...
+      .request('Dashboard', 'getDashboardSetting', { deviceType: this.platformDeviceType })
+      .pipe(
+        take(1),
+        catchError((error) => {
+          if (_.isFunction(error.preventDefault)) {
+            error.preventDefault();
+          }
+          return of(emptyConfig);
+        }),
+        mergeMap((userWidgetConfig: DashboardUserConfig) => {
```
jpspringall, line 103: Complexity here is to allow for backwards compatibility.
Will take whatever is stored in local storage, push it to the server, and then delete it from local storage

vagrant/restart-omv-service.sh
```diff
@@ -0,0 +1,24 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with openmediavault. If not, see <http://www.gnu.org/licenses/>.
+
+set -e
+
+sudo service openmediavault-engined restart
```
jpspringall, line 24: I added this script file to restart omv in vagrant, used by the below

vagrant/sync-usr-share.sh
```diff
@@ -0,0 +1,26 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with openmediavault. If not, see <http://www.gnu.org/licenses/>.
+
+set -e
+
+rsync -av -m /home/vagrant/openmediavault/openmediavault/usr/share/ /usr/share/
```
jpspringall, line 24: This syncs from vagrant share to /usr/share/openmediavault
I used rsync as then it shows me what has changed, to make sure it's expected

deb/openmediavault/workbench/src/app/core/services/dashboard-widget-config.service.ts
```diff
@@ -34,10 +37,16 @@ export class DashboardWidgetConfigService {
 
   private configsSource = new BehaviorSubject<DashboardWidgetConfig[]>([]);
 
+  private get platformDeviceType(): string {
+    return this.platform.ANDROID || this.platform.IOS ? 'mobile' : 'desktop';
```
jpspringall, line 41: This is using CDK See here https://www.decodedfrontend.io/detect-browser-and-os-with-angular-cdk-platform-module/:
It's somewhat limited, but I am using the same device names as from here:
https://www.npmjs.com/package/ngx-device-detector?activeTab=readme

So if it's felt it's too restrictive, the npm version can always the implemented without breaking too much (Hopefully)

deb/openmediavault/debian/openmediavault.dirs
jpspringall: Confirmed this appears to b the case

votdev:
I'll started to review your PR and realized that it only stores the dashboard settings. IMO this does not make sense; what happens with the other settings like hidden/visible datatable columns and other page states that are also stored in the browsers local storage? I can bet that users will complain about that as soon as this PR is published.

I'll refactor and adapt this PR to store all UI settings. I think your work is a good starting point to implement that feature. Thx for all the work done.

votdev:
Superseded by https://github.com/openmediavault/openmediavault/pull/1472.

Thanks for doing the ground work. I think the now implemented solution looks a little bit more lightweight and should make more users happy.

jpspringall:
Hey @votdev,
Sorry for the slow reply, I've been ill for a few days 
No worries, glad to of been of some help, and already had positive feedback :-)

PR #1479, opened 2023-02-01
Add file system creation plugin infrastructure.

votdev:
* Add an infrastructure to allow plugins to append additional file systems to the file system creation page in the UI.
* Enhance the BTRFS file system creation page by adding support of profiles to create various RAID levels.

Peek 2023-02-01 12-24

- [ ] References issue


ryecoaaron (approved):
Seems fine to me.  I don't think it will work for zfs (maybe just a zfs filesystem?) because there needs to be a concept of a pool, volume, and filesystem.

Issue #1241, opened 2022-03-08
Add Snapshots for BTRFS


prk0ghy:
**Is your feature request related to a problem? Please describe.**
I use BTRFS with my openmediavault installation, however it always bothers me, that not many of the features are available via the web ui. As described in the documentation https://openmediavault.readthedocs.io/en/5.x/administration/storage/filesystems.html?highlight=snapshot snapshots are not available.

**Describe the solution you'd like**
I would love to have some way to enable snapshots via webui. It would be great if you can configure snapshots and also access them some way.

**Describe alternatives you've considered**
Snapshots should not replace a good backups but since I want an easy way to roll back changes or get back files this feature would be greatly apprecheated. Configuring from commandline would also be possible but for now I deem it to complicated for comfort.

**Additional context**
Tutorial i found on how to set it up from commandline.
https://linuxhint.com/use-btrfs-snapshots/

votdev:
These features are all planned, but the community seems to have a problem with it: https://github.com/openmediavault/openmediavault/issues/101

paulsijben:
Hopefully to add weight to the ask. 
I am setting up a Pi-based NAS. Tried openmediavault, like it a lot, but the absence of snapshots is an issue for me. So now I am trying Rockstor, which I already like way less, but at least it supports snapshots...

PrplHaz4:
Adding more BTRFS-specific features to OMV would go a long way towards more broad BTRFS adoption.  Some BTRFS features like RAID array management (add, remove, replace, configure) and snapshots are pretty much standard NAS functionality at this point.

IMO, this will also greatly increase the support burden for maintainers, so it would be good to have a plan for enabling the community to take on some of the support aspects.

Issue #312, opened 2019-03-01
Feature Request: Additional BTRFS support in OMV5


flmaxey:
Request additional BTRFS feature support, in OMV5, that equates to some of the features currently available in the ZFS plugin, that supports ZFS.

If sub-volume creation, snapshots, and other BTRFS features are available in the GUI, users may make use them and provide meaningful feedback.

votdev:
IMHO snapshots make only sense with the proposal #101 where shared folders are based on sub-folders. In most cases users create shared folders with sense because they keep special data like photos, videos and so on. It does not make sense to snapshot a whole device that that contains all mixed data. For example if you delete important photos, then you only want to role back the photos shared folder, and not photos, video, documents and so on, because in this case you loose all other data that you won't see too be restored to an older state.

flmaxey:
Understood.
I ran into the same consideration when setting up ZFS;  snapshot the entire pool or segregate specific data types into child file systems, where individual snapshot intervals and settings where possible. And, more importantly, rolling back or creating a clone would apply only to a file system sub-set, not the entire pool.
It would have been nice to look at BTRFS and ZFS side by side, for comparison, but (not being a Dev) I have no idea of the work necessary to make that possible.
Thanks for taking a look.

PelleHanses:
If you want to restore some files from a previous snapshot, you can mount the snapshot under a different name and then just copy the desired files. You do not need to mount the snapshot over the original. This is the case with ZFS anyway and I can't imagine it being handled differently in BTRFS.
Therefore, saving one or all directories in a snapshot doesn't matter much, except you may not want to save unnecessary data in a snapshot.

ulinuxp:
#629 would be great if the btrfs snapshots can also be served with samba.

MarcS1975:
Hi - I saw this conversation and was wondering what the developers update view is on a BTRFS plugin in the OMV gui. Would be nice to have basic management like with ZFS inside the gui. Any thoughts?

claell:
@votdev I don't think I agree with your statement. I think this makes sense at this state already. At least when I correctly understand this proposal. I think it is requesting support for BTRFS features in OVM that are supported for ZFS in OVM.

Those features most likely already make sense (else why would they be availbale for ZFS in the first place).

mi-hol:
#101

mi-hol:
there is no version label but "planned" label.
Will this be worked on for 6.x oder 7.x ?

mi-hol:
[...]

PR #1480, opened 2023-02-02
Issue #1241: Use BTRFS subvolumes for shared folders on BTRFS file systems.

votdev:
Snapshots can be created per shared folder. The snapshot management UI allows to create and delete those snapshots. A restore feature has not been implemented by intention because those shared folders need to be unreferenced from every service beforehand. Instead shared folders should be created from these snapshots, which is a less error-prone approach.

- Subvolumes will be used for shared folders on BTRFS volumes.
- Add snapshot management for shared folders.
- A restore/rollback of snapshots has NOT been implemented by intention because it is difficult to do that when running services access the shared folder. Forcing the users to remove all shares beforehand (as suggested in the docs of a commencial NAS) does not make sense. Instead the user can easily create a shared folder pointing to the snapshot.
- Snapshots are NOT stored below the shared folder, instead they are stored in the root of the BTRFS filesystem in the `.snapshots` subvolume. This is done to do not pollute the shared folder which might cause side-effect, e.g. when syncing data with tools that do not know that `.snapshots` should be ignored.
- Snapshots are named like the shared folder and a timestamp in ISO 8601 format.

```
|- /srv/xxxx (Mount point BTRFS filesystem)
  |- .snapshots
  |- Shared folder 1
  |- Shared folder 2
```

Peek 2023-02-02 18-03

Peek 2023-02-06 18-19

Fixes: https://github.com/openmediavault/openmediavault/issues/312
Fixes: https://github.com/openmediavault/openmediavault/issues/1241

- [x] References issue

### DIFF
--- a/deb/openmediavault/debian/changelog
+++ b/deb/openmediavault/debian/changelog
@@ -4,6 +4,15 @@ openmediavault (6.3.0-1) stable; urgency=low
     file systems to the file system creation page in the UI.
   * Enhance the BTRFS file system creation page by adding support
     of profiles to create various RAID levels.
+  * Issue #1241: Use BTRFS subvolumes for shared folders on BTRFS
+    file systems. Snapshots can be created per shared folder.
+    The snapshot management UI allows to create and delete those
+    snapshots. A restore feature has not been implemented by
+    intention because those shared folders need to be unreferenced
+    from every service beforehand. Instead a shared folders should
+    be created from these snapshots, which is a less error-prone
+    approach. The snapshots are stored in the `.snapshots` folder
+    in the root subvolume of the BTRFS file system.
 
  -- Volker Theile <volker.theile@openmediavault.org>  Tue, 31 Jan 2023 12:31:57 +0100
 
--- a/deb/openmediavault/usr/share/openmediavault/datamodels/rpc.filesystemmgmt.json
+++ b/deb/openmediavault/usr/share/openmediavault/datamodels/rpc.filesystemmgmt.json
@@ -43,7 +43,7 @@
 			},
 			"profile":{
 				"type": "string",
-				"enum": [ "single", "raid0", "raid1", "raid10" ],
+				"enum": [ "single", "dup", "raid0", "raid1", "raid10" ],
 				"required": true
 			},
 			"devicefiles": {
--- a/deb/openmediavault/usr/share/openmediavault/datamodels/rpc.sharemgmt.json
+++ b/deb/openmediavault/usr/share/openmediavault/datamodels/rpc.sharemgmt.json
@@ -253,4 +253,55 @@
 			}
 		}
 	}
+},{
+	"type": "rpc",
+	"id": "rpc.sharemgmt.deletesnapshot",
+	"params": {
+		"type": "object",
+		"properties": {
+			"uuid": {
+				"type": "string",
+				"format": "uuidv4",
+				"required": true
+			},
+			"id": {
+				"type": [ "string", "integer" ],
+				"required": true
+			}
+		}
+	}
+},{
+	"type": "rpc",
+	"id": "rpc.sharemgmt.restoresnapshot",
+	"params": {
+		"type": "object",
+		"properties": {
+			"uuid": {
+				"type": "string",
+				"format": "uuidv4",
+				"required": true
+			},
+			"id": {
+				"type": [ "string", "integer" ],
+				"required": true
+			}
+		}
+	}
+},{
+	"type": "rpc",
+	"id": "rpc.sharemgmt.fromsnapshot",
+	"params": {
+		"type": "object",
+		"properties": {
+			"uuid": {
+				"type": "string",
+				"format": "uuidv4",
+				"required": true
+			},
+			"id": {
+				"type": [ "string", "integer" ],
+				"required": true
+			}
+		}
+	}
 }]
--- a/deb/openmediavault/usr/share/openmediavault/engined/rpc/filesystemmgmt.inc
+++ b/deb/openmediavault/usr/share/openmediavault/engined/rpc/filesystemmgmt.inc
@@ -741,6 +741,8 @@ class OMVRpcServiceFileSystemMgmt extends \OMV\Rpc\ServiceAbstract {
 			}
 			// Create the Btrfs volume. We do not need to create an
 			// partition on each associated device.
+			// https://btrfs.readthedocs.io/en/latest/mkfs.btrfs.html#
+			// https://btrfs.readthedocs.io/en/latest/mkfs.btrfs.html#profiles
 			$cmdArgs = [];
 			$cmdArgs[] = "--force";
 			if (!empty($params['label'])) {
--- a/deb/openmediavault/usr/share/openmediavault/engined/rpc/sharemgmt.inc
+++ b/deb/openmediavault/usr/share/openmediavault/engined/rpc/sharemgmt.inc
@@ -49,6 +49,11 @@ class ShareMgmt extends \OMV\Rpc\ServiceAbstract {
 		$this->registerMethod("getFileACL");
 		$this->registerMethod("setFileACL");
 		$this->registerMethod("getPath");
+		$this->registerMethod("enumerateSnapshots");
+		$this->registerMethod("createSnapshot");
+		$this->registerMethod("deleteSnapshot");
+		$this->registerMethod("restoreSnapshot");
+		$this->registerMethod("fromSnapshot");
 	}
 
 	/**
@@ -132,7 +137,7 @@ class ShareMgmt extends \OMV\Rpc\ServiceAbstract {
 		foreach ($objects as $objectk => $objectv) {
 			$objectDict = new \OMV\Dictionary($objectv->getAssoc());
 			// Set the default values.
-			$objectDict->set("_used", false);
+			$objectDict->set("_used", FALSE);
 			$objectDict->set("device", "");
 			$objectDict->copy("name", "description");
 			$objectDict->set("mntent", [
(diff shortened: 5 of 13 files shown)


PrplHaz4:
This is a great feature and will really allow people to start using the best parts of BTRFS without having to know every detail of the implementation.

There seems to be a lot of churn in the community over snapshot naming being incompatible between some of the popular tools (btrbk, snapper), so it may be worth considering how those will interact.

neheb:
Just tested this. Works great.

paulsijben:
Thank you for adding this!  I have my setup running nicely with btrfs.

One question, how do I automatically prune (too) old snapshots?

rjwonder:
related to the question above, can I just remove the snapshot via command line or will this break somewhere else in the system?

ryecoaaron:
@rjwonder yes

rjwonder:
When having a shared folder on the root of my drive /srv/dev-disk-by-uuid-xxx/ I can snapshot the full thing. 'btrfs subvolume list <path>' is showing only  .snapshots as subvolume. This works well for me as I like to feed the snapshot to snapraid (https://github.com/automorphism88/snapraid-btrfs). 

When there used to be shared folders (now deleted), the sub volumes seems to stay. I think preventing me from snapping the full drive. When trying to setup a new share on /srv/dev-disk-by-uuid-xxx/ the snapshot creates an entry in .snapshots and all folders, but no files. I guess this has to do with BTRFS and the facts that snapshots never go into child sub volumes? 

Would this be the best way to do full disk snapshots? Use CoW to copy the data to a different folder. Use delete subvolume on original (includes data deletion) and create a shared folder on  /srv/dev-disk-by-uuid-xxx/? If I want shared folders on specific folders (e.g. for samba) then put them on the pool which is not BTRFS for OMV?

votdev:
> related to the question above, can I just remove the snapshot via command line or will this break somewhere else in the system?

Yes, OMV is not doing any magic here.

MarcS1975:
So the creation of every 'Shared Folder' on a BTFRS drive will create a "Subvolume" and therefore be ignored by a Snapraid sync...

votdev:
> So the creation of every 'Shared Folder' on a BTFRS drive will create a "Subvolume" and therefore be ignored by a Snapraid sync...

Snapraid is no official plugin of OMV, so i only can enhance [the documentation](https://github.com/openmediavault/openmediavault-docs/commit/4686155e697e70a70e4b52e6530c2acc371f942a).

votdev:
> So the creation of every 'Shared Folder' on a BTFRS drive will create a "Subvolume" and therefore be ignored by a Snapraid sync...

There is a workaround for that situation. If the directory already exists, the shared folder creation mechanism does not create a new subvolume when you point to that directory in the shared folder creation dialog in the UI. The UI gives you more or less already that hint. This way you can create shared folders with classic directories on BTRFS file systems.

ryecoaaron:
As I posted on the forum, until snapraid supports btrfs subvolumes, there is nothing I can do with the snapraid plugin to help.

MarcS1975:
As posted on the Forum, a simple flag ON/OFF in the Plugin would do the job and satisfy both Snapraid users and BTRFS Fans

votdev:
> As I posted on the forum, until snapraid supports btrfs subvolumes, there is nothing I can do with the snapraid plugin to help.

It's on snapraid to support subvolumes. Hey, these are regular directories from the file system consumer point of view. Everyone can access and process them as normal. So why ignoring them?

MarcS1975:
Subvolumes are not filesystems. They actually just act like filesystems but
are on a higher level of abstraction.

Btrfs wiki "A Btrfs subvolume can be thought of as a separate POSIX file
namespace, mountable separately by passing subvol or subvolid options to
the mount(8) utility."

Snapraid ignores Snapshots because of data integrity issues and my guess is
SR will never support it. But this would go to far for this discussion here.

Bottom line you decide to go down the Subvolume route it excludes Snapraid
use. But thats your free choice of course.

MarcS1975:
I commend your enthusiasm to focus on one filesystem and add more Plugin functionality for OMV, but what I used to love about OMV was always the great flexibility to freely choose filesystems and RAID management solutions. Now it all seems to go into BTRFS support as a filesystem, and especially as a RAID solution.

**I assume you are aware of the endless issues of BTRFS. All the unfinished stuff, the severe RAID5/6 issues and the messy implementation of BTRFS storage namespaces.** 

https://arstechnica.com/gadgets/2021/09/examining-btrfs-linuxs-perpetually-half-finished-filesystem/

`The btrfs-raid5 and btrfs-raid6 topologies are extremely unreliable. The btrfs wiki itself describes btrfs parity RAID as "[mostly implemented](https://btrfs.wiki.kernel.org/index.php/FAQ#Can_I_use_RAID.5B56.5D_on_my_Btrfs_filesystem.3F)," and it explicitly recommends "[btrfs] parity RAID [should] be used only for testing purposes."`

https://www.phoronix.com/news/Btrfs-Warning-RAID5-RAID6
It remains impossible to use BTRFS as RAID 5/6 solution.

**The issues are so severe that the biggest NAS company Synology decided to keep BTRFS as a filesystem but decided to use Linux RAID (!) on its systems.**
https://kb.synology.com/en-uk/DSM/tutorial/What_was_the_RAID_implementation_for_Btrfs_File_System_on_SynologyNAS

ryecoaaron:
OMV doesn't even allow you to create raid5/6 with btrfs but things are getting better - https://www.phoronix.com/news/Linux-6.2-Btrfs-EXT4

> but what I used to love about OMV was always the great flexibility to freely choose filesystems and RAID management solutions. Now it all seems to go into BTRFS support as a filesystem, and especially as a RAID solution.

Nothing has changed with ext4 and xfs.  You can still use mdadm.  So, you can still freely choose.  The snapshot stuff can't be implemented for the filesystems because they don't support it.

MarcS1975:
> things are getting better - https://www.phoronix.com/news/Linux-6.2-Btrfs-EXT4
That's just Marketing. Thigs are still **not** getting better say users:
https://www.reddit.com/r/btrfs/comments/127zmsx/what_do_you_think_about_the_kernel_62_btrfs/




> OMV doesn't even allow you to create raid5/6 with btrfs
Well some of your Moderators in the OMV Forum don't know this. They advise that "BTRFS replaces SnapRaid. "

ryecoaaron:
The first user in your reddit thread says

votdev:
> I commend your enthusiasm to focus on one filesystem and add more Plugin functionality for OMV, but what I used to love about OMV was always the great flexibility to freely choose filesystems and RAID management solutions. Now it all seems to go into BTRFS support as a filesystem, and especially as a RAID solution.

This is wrong. You can still use whatever file system you want. The only thing is that OMV fully integrates Btrfs feature support into the UI. Nothing more. You can still choose. That snapraid does not handle subvolumes correct is a bad situation, but there are workflows which will enable you to still use it.

MarcS1975:
> What is the goal of all of this for you? Do you want all btrfs removed?

Of course not! Nothing shall be removed. Its a filesystem option for anyone to try out. I use it myself as a filesystem (not as RAID). So it will always require some kind of RAID on top of BTRFS. And my surprise was just, that the the new OMV functionality makes it impossible to use some popular RAID solutions. So my "Goal" _was_ to find a compromise (the tick box), so users can continue to use their **legacy** OMV systems based on Snapraid. I understand that this is not going to happen, and we have to live with 'workarounds'. Fine.
So then my '2nd Goal' is to learn and understand what brought you guys to favour BTRFS so much, when it is so widely dismissed as an unfinished and unreliable solution. If you say, you disagree and not interested in sharing reasons then that also fine.
:)

votdev:
> So then my '2nd Goal' is to learn and understand what brought you guys to favour BTRFS so much, when it is so widely dismissed as an unfinished and unreliable solution. If you say, you disagree and not interested in sharing reasons then that also fine. :)

ZFS will never go into the Linux kernel, so Btrfs is the way to go. If you prevent RAID5 and RAID6, then you should be fine. Btrfs is the default file system in many Linux distributions nowadays, so it might not be too bad. Second, the main feature i wanted to have in OMV was the ability to create snapshots of shared folders to allow easy recovering of previous states. This is super easy with Btrfs. I don't think there must be more reasons to switch to Btrfs.

ryecoaaron:
Personally, I only favor btrfs over zfs. And that is only because btrfs is easier to support codewise and people wise. my first choice in filesystems is still ext4 but I use ext4,xfs, and btrfs with my backups just in case a bad bug hits one of the filesystems. I dont run any kind of raid anymore and prefer borgbackup to snapraid or other bitrot detection systems.

MarcS1975:
> i wanted to have in OMV was the ability to create snapshots of shared folders to allow easy recovering of previous states. 
> This is super easy with Btrfs.
Agree! And that's the only reason I use BTRFS, even at the cost of having to put up with a RAID on top of it.

MarcS1975:
> I dont run any kind of raid anymore and prefer borgbackup to snapraid or other bitrot detection systems.
I see. Have you ever compared the space requirements of borgbackup vs snapraid? Borg is very efficient but my assumption is that Borg still use a lot more space than Snapraid's single parity.

ryecoaaron:
> Have you ever compared the space requirements of borgbackup vs snapraid? Borg is very efficient but my assumption is that Borg still use a lot more space than Snapraid's single parity.

No.  But on a very important backup, all of the source files (sql dumps, images, source code, etc) consume about 383GB of space.  The borgbackup directory with 26 backups (8 hourly, 7 daily, 4 weekly, 5 monthly, 2 yearly) consumes 142 GB using dedupe and compression.  This same setup is running on three servers with two disks each.  It gives me the ability to restore from 26 different times easily just by picking an archive.

MarcS1975:
So if a disk fails, instead of replacing a disk and letting the RAID do the resilvering work, you replace the disk and then manually select the archives to restore to the new disk?

ryecoaaron:
No.  All three servers are backup servers.  Being highly available is not important for the backup servers.  If a disk on a backup server with the primary backup (rsync'd from the primary server) fails, I just replace the disk and the scheduled rsync will fix it.  If the disk with the borgbackup fails, I would rsync the borgbackup from one of the other backup servers to it.  The primary server is running two nvme sticks in raid 1.  I also have offsite backups.

Issue #1500, opened 2023-02-20
Add settings to alter snapshot naming


Manfred-Knick:
**Is your feature request related to a problem? Please describe.**
BACKGROUND:
Exporting /srv/dev-disk-by-uuid-...../.snapshots as a SMB/CIFS share,
Windows is not able to digest the two ":" colons used in the ISO 8601 time part of the snapshot name,
being a "not-allowed" character in Windows file naming conventions.
Instead, the Windows Explorer substitutes a crippled random replacement for display.
Furtheron, all snapshots are displayed with one identical date of change = date of creation.

==> The Windows user is not able to identify the correct snapshot he is searching for.

**Describe the solution you'd like**
Multiple selectable options, or a configurable option
E.g.:
Changing the ":" separator to (e.g.) "-" like in the date part would circumvent this (M$) problem.
Another alternative configuration could be to omit "-" and ":" completely, resulting in a name ending as
. . . . . _yyyymmddThhmmss

**Describe alternatives you've considered**
To me, "File-name character translation" files referred-to by registry elements 
do not shine as a viable workaround.

**Additional context**
 Invalid filenames on Windows:
    The filename doesn’t contain any of the following characters: 
" (double quote), 
* (asterisk), 
* < (less than), 
* > (greater than), 
* ? (question mark), 
* \ (backslash), 
* | (pipe), 
* / (forward slash), 
* : (colon) <-------------------------------------!

votdev:
> Changing the ":" separator to (e.g.) "-" like in the date part would circumvent this (M$) problem

This is not supported by ISO 8601. I will not use a naming scheme that is not standardized.

> Another alternative configuration could be to omit "-" and ":" completely, resulting in a 
> name ending as . . . . . _yyyymmddThhmmss

This is the ISO 8601 basic format. Looks a little bit weird and makes the date/time really unreadable, but if Windows does not like it, there is no alternative.

PR #1501, opened 2023-02-20
Issue #1500: Use ISO 8601 basic format in the snapshot naming schema.

votdev:
Switch from ISO 8601 extended to basic format because SMB/Windows does not support `:` in file names. Note that due to this change, earlier snapshots with the ISO 8601 extended format are no longer listed under Windows in `Previous Versions`.

The naming schema is hardcoded by intention because PHP and date (CLI) do not use the same pattern which would result in using two environment variables and completely denies the use of a UI setting to allow users to specify their own naming schema. IMO, OMV does not need to support everything. Following KISS is sometimes the better choice.

Fixes: https://github.com/openmediavault/openmediavault/issues/1500

- [x] References issue

### DIFF
--- a/deb/openmediavault/debian/changelog
+++ b/deb/openmediavault/debian/changelog
@@ -2,6 +2,11 @@ openmediavault (6.3.2-1) stable; urgency=low
 
   * Issue #1490: Support SMB shadow copies for home directories
     when the shared folder is located on a BTRFS file system.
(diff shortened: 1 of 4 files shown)


Manfred-Knick:
That was quick!
Thanks a lot!

> IMO, OMV does not need to support everything.

I deeply agree!

> Following KISS is sometimes the better choice.

Always!

Kind regards
Manfred

PR #1507, opened 2023-03-01
Add the ability to specify the metric (relative priority) …

votdev:
… of the default route set via the `Gateway` option in the network interface page.

The options gateway4, gateway6 will become deprecated in next Netplan version. By setting the default route the new way we will also get the opportunity to add support for custom `metric` settings.

- https://manpages.debian.org/net-tools/route.8.en.html
- https://netplan.readthedocs.io/en/stable/netplan-yaml/#properties-for-all-device-types
- https://netplan.readthedocs.io/en/stable/netplan-yaml/#default-routes

- [ ] References issue


ryecoaaron (approved):
Your changes are very similar to the changes I have been making for Ubuntu 22.04.  So, I think it looks good.

Issue #1516, opened 2023-03-21
Allow Workbench to be reverse proxied through subfolder


EmTeedee:
**Is your feature request related to a problem? Please describe.**
I want to use a reverse proxy to access OMV through a subfolder of my main domain e.g. https://example.com/omv/

**Describe the solution you'd like**
All dependencies linked in the web ui should be using relative paths.

**Describe alternatives you've considered**
Using a separate subdomain would be an alternative, the solution would give more flexibility for users.

votdev:
Are you sure you are fixing the issue at the right side?

votdev:
# Testing setup

Create the file `Caddyfile` with the following content:
```
{
  debug
  admin off
}
:8080 {
  route /foo/* {
    uri strip_prefix /foo
    reverse_proxy http://<OMV_IP_ADDR>
  }
  redir /foo /foo/
}
```

Start the proxy via `podman run --replace -it --name=omv-via-proxy -v ./Caddyfile:/etc/caddy/Caddyfile:Z -p 8080:8080 docker.io/library/caddy:latest`.

The UI can now be reached via `http://localhost:8080/foo`.

PR #1517, opened 2023-03-21
Change absolute asset paths in workbench to relative

EmTeedee:
Changes absolute asset paths in workbench to relative. Almost all other paths are already relative.

Fixes: #1516 

Signed-off-by: Tobias Tiederle <git.emteedee@das-komitee.de>

- [x] References issue

### DIFF
--- a/deb/openmediavault/workbench/src/app/core/components/navigation-bar/navigation-bar.component.scss
+++ b/deb/openmediavault/workbench/src/app/core/components/navigation-bar/navigation-bar.component.scss
@@ -15,7 +15,7 @@
     .logo {
       width: 100%;
       height: 100%;
-      background-image: url('/assets/images/openmediavault_logo_simple.svg');
+      background-image: url('../../../../assets/images/openmediavault_logo_simple.svg');
       background-repeat: no-repeat;
       background-size: contain;
       background-position-y: center;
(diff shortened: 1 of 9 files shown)


votdev:
Can you confirm that this is working in your setup?

EmTeedee:
Sorry, I had problems setting up the build environment.
I see the same issue, working on resolving it.

I've changed the paths in the compiled webpack and it's working, just need to get the sass compiler to output the same.

I'll tag you again when I get that working.

votdev:
Have you ever checked how the paths are looking after this PR has been applied. Some are in the root like `/foo.svg`, others look like `./assets/images/bar.gif`. This does not really look like a valid improvement, this is IMO spaghetti code. I am sorry, but i can not approve and merge it this way.

votdev (changes requested):
The PR needs an overall refactoring to do not result in spaghetti code where some images are located in the root, others are still in `/assets/xxx`.

EmTeedee:
Thanks for fixing the relative path issue.

I stayed away from the `^` because of this comment from the angular project:
>  [...] paths starting with ^ is a Webpack specific feature and as such we don't provide support nor guarantee that these will continue to work in the future.
https://github.com/angular/angular-cli/issues/18013#issuecomment-649373940

Either way, it works and when it stops they hopefully provide a supported way of achieving the same, angular/scss just doesn't at the moment.

votdev:
> I stayed away from the `^` because of this comment from the angular project:

There is no alternative to that right at the moment. At least i do not know about or have found any.

Issue #1410, opened 2022-10-21
Add "--base" option in WeTTY plugin


kirtapix:
Hello there,

I'm adding extra location conf files to nginx because i want to reach web services by specifying a subfolder and not a port in the url 
(my brain can't remember numbers, but words are ok ahah) 
Here is an example for WeTTY :
127.0.0.1:2222 -> 127.0.0.1/wetty
The problem is I can't specify the base subfolder option for wetty, it's hardcoded in the serviced file.

Is it possible to add this option in the plugin management web panel ?
![image](https://user-images.githubusercontent.com/37413390/197180270-6b1941cf-e438-487d-925a-6be37fc53526.png)

Thanks in advance.

PR #1518, opened 2023-03-22
add base path option to wetty

EmTeedee:
Fixes: #1410

Built and tested.
When using a reverse-proxy setup the "Open UI" button might not work, depending on hostname and port of the proxy. To address that, one would need an extra option for that just for this one button.

- [x] References issue


votdev:
Please introduce a new environment variable Adding such an option in the UI breaks the philosophy of OMV.

votdev:
> When using a reverse-proxy setup the "Open UI" button might not work, depending on hostname and port of the proxy. To address that, one would need an extra option for that just for this one button.

That might be a problem if there is an environment variable used. But TBH, this is a special edge case which is not the main purpose of OMV. I say it as often as needed, OMV is no Webmin where the UI provides whatever possible option is available for a service.

votdev (changes requested):
Isn't that doable via path rewriting rules in the proxy?

EmTeedee:
Wetty provides the base option to address the problem, if openmediavault has something similar, I haven't found it.

votdev:
> Wetty provides the base option to address the problem, if openmediavault has something similar, I haven't found it.

That's because Angular sets that during compilation and changing `<base>` tag at runtime is a little bit hackish.

PR #1541, opened 2023-05-07
Upgrade russian translate

Psychosynthesis:
Add and correct the translation. I left some lines untranslated, as they are not unambiguous (it can be translated in different ways). However, there are few such places, so I think it's acceptable.

I checked everything, but since there are a lot of lines, purely theoretically, I could skip double quotes somewhere or add extra ones. I don't know how to build the project, so I can't fully check this. It would be great if one of the contributors did this before the merge.

- [ ] References issue


votdev (approved):
normally the translations are managed by Transifex and only pulled into the Git repository. Please check the following URLs if you want to contribute more translations.

- https://translate.openmediavault.org
- https://docs.openmediavault.org/en/stable/development/contribute.html#how-to-become-a-translator

PR #1553, opened 2023-05-16
Refactor systemd-networkd Salt state

votdev:
Relocate all Netplan.io related Salt states into the sub-directory called `netplan`. This will allow us to easily establish the symlink between `/etc/resolv.conf` and `/run/systemd/resolve/resolv.conf` AFTER the Netplan configuration has been applied.

- [ ] References issue


votdev:
@ryecoaaron I hope this will not affect any plugin.

Issue #1564, opened 2023-07-12
Support forced/custom Mac Addresses from the network interfaces tab


Crypto-Spartan:
**Is your feature request related to a problem? Please describe.**
For better use of interface management, I would like to be able to force my own MAC Addresses, particularly on VLANs. 

**Describe the solution you'd like**
On the webpage where you are able to modify an interface, have an additional field that says something like `Forced MAC Address` and allows the user to type in whatever MAC Address they want to use for that interface.

**Describe alternatives you've considered**
While I was looking into either modifying the salt files or adding my own networkd-systemd config, I realized that this wouldn't be terribly hard to build into OMV. I would rather have it as a built-in feature than have a hacky solution.

**Additional context**
I'm currently working on a PR now.

PR #1565, opened 2023-07-13
Add alternative MAC address support for wired, wireless and VLAN network interfaces

Crypto-Spartan:
Support forced/custom Mac Addresses from the network interfaces tab

This PR adds support for custom/forced MAC addresses on Ethernet, VLAN, & Wifi interface types. It is implemented similar to the DNS fields, if the field is blank on the webpage, no custom value is input into the `{{ 20 | 30 | 50 }}-openmediavault-{{ interface.devicename }}.yaml` file that gets generated for the network configuration.

Fixes: #1564


votdev (approved):
@Crypto-Spartan Thx for your support.

PR #1575, opened 2023-08-10
[Workbench] Changing forms from page-width to a more reasonable size

gutiar-junky:
Changed every form in the angular workbench to a more reasonable size.

By adding some global classes and extending the FormPageConfig-Interface, there's now the possibilty to style each mat-card with one or more css classes.

I really don't want to do these CLA things - just copy my code if you want to.

### DIFF
--- a/deb/openmediavault/workbench/src/app/core/components/intuition/models/form-page-config.type.ts
+++ b/deb/openmediavault/workbench/src/app/core/components/intuition/models/form-page-config.type.ts
@@ -98,6 +98,7 @@ export type FormPageConfig = {
   // Form pages can have buttons in their footer.
   buttonAlign?: 'start' | 'center' | 'end';
   buttons?: Array<FormPageButtonConfig>;
+  styling?: FormPageStylingConfig;
 };
 
 export type FormPageButtonConfig = {
@@ -173,3 +174,7 @@ export type FormPageButtonExecute = {
   // A callback function. Internal only.
   click?: (buttonConfig: FormPageButtonConfig, values: FormValues) => void;
 };
+
+export interface FormPageStylingConfig {
+  cardClasses: string;
+}
--- a/deb/openmediavault/workbench/src/app/core/components/intuition/models/form-page-config.type.ts
+++ b/deb/openmediavault/workbench/src/app/core/components/intuition/models/form-page-config.type.ts
@@ -175,6 +175,6 @@ export type FormPageButtonExecute = {
   click?: (buttonConfig: FormPageButtonConfig, values: FormValues) => void;
 };
 
-export interface FormPageStylingConfig {
+export type FormPageStylingConfig {
   cardClasses: string;
 }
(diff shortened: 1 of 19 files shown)


deb/openmediavault/workbench/src/app/core/components/intuition/models/form-page-config.type.ts
```diff
@@ -173,3 +174,7 @@ export type FormPageButtonExecute = {
   // A callback function. Internal only.
   click?: (buttonConfig: FormPageButtonConfig, values: FormValues) => void;
 };
+
+export interface FormPageStylingConfig {
```
votdev, line 178: OMV is using `type` allover.

deb/openmediavault/workbench/src/styles.scss
```diff
@@ -99,6 +99,20 @@ $omv-dark-theme: mat.define-dark-theme(
   width: auto;
 }
 
+.mat-card.omv-w-xs {
```
votdev, line 102: IMO this could be done more nicer with CSS `@media` rule.

deb/openmediavault/workbench/src/app/core/pages/login-page/login-page.component.ts
```diff
@@ -76,7 +76,10 @@ export class LoginPageComponent implements OnInit {
           click: this.onLogin.bind(this)
         }
       }
-    ]
+    ],
+    styling: {
+      cardClasses: 'test'
```
votdev, line 81: ???

deb/openmediavault/workbench/src/app/pages/diagnostics/system-logs/system-logs-remote-form-page.component.ts
```diff
@@ -89,6 +89,9 @@ export class SystemLogsRemoteFormPageComponent extends BaseFormPageComponent {
           url: '/diagnostics/system-logs'
         }
       }
-    ]
+    ],
+    styling: {
+      cardClasses: 'omv-w-xs'
```
votdev, line 94: Not the best solution to have the need to touch every form page to specify the styling.

deb/openmediavault/workbench/src/styles.scss
gutiar-junky: i think, the width is not depending on the screen-width at all, so @media is not really appropriate.
The width of the form should be dependent by it's contents.

deb/openmediavault/workbench/src/app/core/pages/login-page/login-page.component.ts
gutiar-junky: missed to remove - sorry!

deb/openmediavault/workbench/src/app/pages/diagnostics/system-logs/system-logs-remote-form-page.component.ts
gutiar-junky: the other way 'round would be to make every form smaller by default, see https://github.com/openmediavault/openmediavault/pull/1575#discussion_r1290175835

votdev:
I will close this because i do not like the approach. I've a better and more simpler in mind. Thank you for your commitment. It has at least been a good occasion to think about the issue again.

votdev:
Superseded by https://github.com/openmediavault/openmediavault/pull/1577

Issue #1596, opened 2023-09-18
improve ethernet interface detection


billzez:
Linux has switched from using virtual vlan interfaces for switch ports to "dsa" which creates physical ethernet interfaces that represent the true state of the switch port.  These ports are named by the device-tree and there is no convention for them, and since OMV detects interfaces by name, these ports are ignored.

Two suggested options:
1.  Treat DSA ports as normal ethernet ports.  If /sys/class/net/*iface*/type is "1" (ARPHDR_ETHER - see if_arp.h), the device is ethernet.
2.  Treat DSA ports as special.  Can use "udevadm info /sys/class/net/*iface*" and look for "DEVTYPE=dsa"

As second problem is that DSA masters should be ignored.  If "/sys/class/net/*iface*/dsa/", the ethernet interface is a DSA master.

https://github.com/torvalds/linux/blob/master/include/uapi/linux/if_arp.h
https://www.kernel.org/doc/Documentation/networking/dsa/dsa.txt

votdev:
DSA is for router and switches. What do you expect from this on a NAS?

billzez:
We created an OMV image for this board:

https://www.axzez.com/product-page/interceptor-carrier-board-with-cm4102032

Discussed here:

https://forum.openmediavault.org/index.php?thread/41469-axzez-inteceptor/&postID=366621#post366621

votdev:
Send me an hardware example and I can try to get it running in OMV. Without being able to test such things it will be hard to do that.

ryecoaaron:
Possible working netplan file:
`aaron@interceptor:/etc$ cat netplan/10eth0.yaml 
network:
  version: 2
  renderer: networkd
  ethernets:
    wan@eth0:
      dhcp4: yes
      dhcp4-overrides:
        use-dns: false
        use-domains: false
      dhcp6: no
      link-local: []
      nameservers:
        addresses:
        - 192.16.1.1
        search: [maxwellpreston.com]
    lan0@eth0:
      dhcp4: yes
      dhcp6: no
    lan1@eth0:
      dhcp4: yes
      dhcp6: no
    lan2@eth0:
      dhcp4: yes
      dhcp6: no
    ext1@eth0:
      dhcp4: yes
      dhcp6: no
    ext2@eth0:
      dhcp4: yes
      dhcp6: no`

output of ip a
`aaron@interceptor:/etc$ ip a
[...]

PR #1600, opened 2023-09-26
Issue #1596: Support DSA network interfaces

votdev:
Fixes: https://github.com/openmediavault/openmediavault/issues/1596

- [x] References issue


ryecoaaron:
omv-firstaid shows eth0 and the dsa interfaces.
image
Creating an interface in the web interface is adding entries to the dropdown but there is no text.
image

votdev:
> omv-firstaid shows eth0 and the dsa interfaces. image Creating an interface in the web interface is adding entries to the dropdown but there is no text. image

OK, the `eth0` was expected, but not that parts of the interface names are stripped or no text is displayed. Need to find a way to emulate the hardware what suxx.

PR #1613, opened 2023-10-14
Update German translation

WebSnke:


votdev:
Thanks for helping OMV to translate the UI. Please create an account at Transifex and do the translations there. Translations are sync'd from Transifex to GitHub and not the other direction. Because of that all changes by this PR will get overwritten with the next sync.

Please check the docs [how to become a translator](https://docs.openmediavault.org/en/stable/development/contribute.html#how-to-become-a-translator).

PR #1617, opened 2023-10-20
Add suggestions to `textInput` form field

votdev:
Peek 2023-10-21 10-27

See https://v14.material.angular.io/components/autocomplete/overview


votdev:
@ryecoaaron This might be interested for your plugins as well.

ryecoaaron:
I can think of a couple of places this would be useful. Thanks!

PR #1618, opened 2023-10-26
ClamAV cannot start - AppArmor blocks access to log file

rabesocke:
Fix a problem with accessing the log file that prevents ClamAV to start

Solution found here: https://www.mail-archive.com/clamav-users@lists.clamav.net/msg50435.html


votdev:
Thx for the fix. Can you please modify the commit message and add `Signed-off-by: Foo Bar <foo.bar@test.com>` so that GH does not block the PR merging.

rabesocke:
I will close this PR and open a new one with signoff. I did a mistake with changing the commit message of this commit / PR.

New PR: #1620 
@votdev

Issue #1622, opened 2023-10-28
Add SD-Card Support to openmediavault-usbbackup Plugin


HeyMeco:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Being able to select /dev/mmcblkX as a device to be mounted, copy the files to selected shared folder and unmounted.

**Describe alternatives you've considered**
A systemd script but it would not be as clean and integrated.

**Additional context**
As someone who handles lots of SD Cards I would love to just connect them to my OMV Server and them being ingested to a shared folder in which a folder named by the current date is the copy location.

Maybe usbbackup can become "externalbackup"

votdev:
Why not using SD card USB adapter?

HeyMeco:
@votdev > Why not using SD card USB adapter?

As many devices especially Single Board Computers already have a Reader it would make sense to be able to use the capabilities which are already there.

votdev:
If the community contributes the feature i am fine with that.

superenginegit:
I'd like to contribute this, if no one is working on it

PR #1630, opened 2023-11-15
Accept mmcblk devices for usbbackup

superenginegit:
The changes in this PR allow SD-Cards and other mmcblk devices to be selected for the usb-backup plugin.

Fixes: [https://github.com/openmediavault/openmediavault/issues/1622](https://github.com/openmediavault/openmediavault/issues/1622)


deb/openmediavault/usr/share/php/openmediavault/system/storage/storagedevice.inc
```diff
@@ -200,6 +200,32 @@ class StorageDevice extends \OMV\System\BlockDevice
...
 
+	/**
+	 * Check if the given device is an SD or MMC Device.
+	 * @return boolean true if device is SD or MMC, otherwise FALSE.
+	 */
+	public function isMmcblk() {
+		// Try to identify SD card by 'ID_PATH=xxx.mmc'
+		// Example (raspberry-pi):
+		// ID_PATH=platform-3f202000.mmc
+		if (TRUE === $this->hasUdevProperty("ID_PATH")) {
```
votdev, line 211: Is this code block really necessary. Are there devices that are not named ´/dev/mmcxxx´?

deb/openmediavault/usr/share/php/openmediavault/system/storage/storagedeviceinterface.inc
```diff
@@ -181,6 +181,12 @@ interface StorageDeviceInterface {
 	 */
 	public function isAta();
 
+	/**
+	 * Check if the given device is an SD or MMC Device.
```
votdev, line 185:
```suggestion
	 * Check if the given device is an SD or MMC device.
```

votdev (changes requested):

deb/openmediavault/usr/share/php/openmediavault/system/storage/storagedevice.inc
superenginegit: No, I've removed the block.

HeyMeco:
Does this change also correctly unmount the SD-Card?

superenginegit:
It is done exactly the same as for any usb device before. So it should do that.

votdev:
@superenginegit Could you please do a test if my changes are working.

I'm not happy by adding a new interface method (`isMmc`) that nothing has to do with the general interface. I would like to see something like `isKindOf("USB")`, isKindOf("MMC") or something else that is generic and can be reused by future improvements. But this requires a refactoring of the storage backend classes which makes the code incompatible for all OMV6 plugins. The current approach is therefore the one that causes the fewest problems.

superenginegit:
I've tested it, and it works on my testing setup.

votdev:
The feature will be available in OMV7.

HeyMeco:
I just tested it and I'm not sure why but it didn't unmount. Also I can't select the device if there isn't also an SD-Card connected already. I made this shell script that maybe demonstrates what I meant better: https://github.com/HeyMeco/sd-copy . It would also be great if we could have a new folder with date/timestamp as name for every time the sync runs.

votdev:
> I just tested it and I'm not sure why but it didn't unmount. 

Would be good if you could find out why this is happening. Other users might benefit from that contribution.

> Also I can't select the device if there isn't also an SD-Card connected already.

That is because the kernel does not see any device file; so OMV can not display it as well.

> It would also be great if we could have a new folder with date/timestamp as name for every time the sync runs.

This is not the purpose of this plugin. The main purpose is to sync new/removed files between a shared folder and an external device. What you have in mind is a backup feature that has nothing to do with the main purpose of the plugin nor the underlying rsync tool it is based on.

PR #1647, opened 2023-12-08
Add omv-release-upgrade infrastructure

votdev:
Add infrastructure to allow plugins to customize the distribution upgrade from OMV6/Debian11 to OMV7/Debian12.


ryecoaaron (approved):
Looks ok to me. I need to add the components to omv-extras now.

PR #1650, opened 2023-12-14
Add UTF8 support for SSL certificates created in Web UI

MestreLion:
By default SSL certificates created with `openssl req -x509` do not properly decode UTF8 strings in the subject (Organization Name, Unit, City, etc), and are not correctly displayed in "Show Certificate" UIs in browsers and other GUI tools (although they are properly displayed in OMV's Web GUI)

Add `-utf8` command-line option to `openssl` so all clients properly interpret the UTF8 strings in the certificate.

### DIFF
--- a/deb/openmediavault/usr/share/openmediavault/engined/rpc/certificatemgmt.inc
+++ b/deb/openmediavault/usr/share/openmediavault/engined/rpc/certificatemgmt.inc
@@ -320,6 +320,7 @@ class CertificateMgmt extends \OMV\Rpc\ServiceAbstract {
 		$cmdArgs[] = "-nodes";
 		$cmdArgs[] = sprintf("-days %d", $params['days']);
 		$cmdArgs[] = "-sha256";
+		$cmdArgs[] = "-utf8";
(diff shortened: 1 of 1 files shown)


votdev:
Thx

votdev:
49e7eba664c080afe057cc272311adecfcf6baaa is releated to this feature.

MestreLion:
> [49e7eba](https://github.com/openmediavault/openmediavault/commit/49e7eba664c080afe057cc272311adecfcf6baaa) is releated to this feature.

Yes, they're complementary but independent settings:
- `-utf8` when _creating_ the certificate is _necessary_ for tools and GUIs such as FireFox's and Chrome's "Show Certificate", Gnome's "Certificate Viewer", etc, to properly display the subject, otherwise it treats the UTF8 as Windows-1252 ("ANSI") and show garbled characters. 
- `-nameopt utf8` is useful when _displaying_ the certificate as text, as by default `openssl -text` shows any non-ASCII bytes as hex-escaped  (`\xC3` etc) even for certificates created with `-utf8`

Issue #1652, opened 2023-12-15
Replace cron-apt by unattended-upgrade


votdev:
It looks like [unattended-upgrade](https://wiki.debian.org/UnattendedUpgrades) is becoming as powerful as cron-apt. Maybe it is time to change.

PR #1653, opened 2023-12-16
Use the unattended-upgrades package instead on cron-apt

votdev:
Fixes: https://github.com/openmediavault/openmediavault/issues/1652


ryecoaaron:
You might want to look at setting Unattended-Upgrade::Allowed-Origins

votdev:
> You might want to look at setting Unattended-Upgrade::Allowed-Origins

I can not find that entry in /etc/apt/apt.conf.d/50unattended-upgrades. Is this still available in Bookwork? What do you suggest to set there?

votdev:
I just want to auto-install security updates only.

ryecoaaron:
Then you would just want:
Unattended-Upgrade::Allowed-Origins {
"${distro_id}:${distro_codename}-security";
};

PR #1655, opened 2023-12-17
Add email to SSL certificate's SAN in its own field

MestreLion:
Currently, email is added to subjectAltName in the DNS field (from commonName), as `DNS:hostname/emailAddress=x@a.com`, which renders as a single field, as if the email was part of the hostname. While accepted by Firefox (and possibly others), this is not the correct syntax.

This PR changes the SAN to `DNS:hostname,email:copy`, which properly copies the emailAddress value from the subject and renders as `DNS:hostname, email:x@a.com` (2 distinct fields).

Note 1: Current codebase seems to use both `<TAB>` and `  ` (2 spaces) as alignment for continuation lines (after indentation). Since recent commits seems to favor `<TAB>`, I've used that in the PR

Note 2: Isn't it supposed to have a Comment field in the SSL certificate creation form? I can see it in the code (both in the HTML template and in the form processing code), and I also noticed a recent commit changing it from Text Input to Tag Input, but I don't see the field anywhere in the UI form (using latest OMV6): 
image

### DIFF
--- a/deb/openmediavault/usr/share/openmediavault/engined/rpc/certificatemgmt.inc
+++ b/deb/openmediavault/usr/share/openmediavault/engined/rpc/certificatemgmt.inc
@@ -281,7 +281,7 @@ class CertificateMgmt extends \OMV\Rpc\ServiceAbstract {
 		// Validate the parameters of the RPC service method.
 		$this->validateMethodParams($params, "rpc.certificatemgmt.create");
 		// Generate the certificate subject string.
-		$rdnFields = [ "c", "st", "l", "o", "ou", "cn" ];
+		$rdnFields = [ "c", "st", "l", "o", "ou", "cn", "email" ];
 		$subject = "";
 		foreach ($rdnFields as $rdnFieldk => $rdnFieldv) {
 			if (empty($params[$rdnFieldv]))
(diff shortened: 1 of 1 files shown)


votdev:
Do you have any links to a documentation about this issue?

MestreLion:
> Do you have any links to a documentation about this issue?

All tutorials/QAs and the like  I've seen so far on how to create a self-signed SSL certificate use the format `subjectAltName=DNS:xxxx,email:a@b.com`, (I've never seen _any_ mention of format `subjectAltName=DNS:xxxx/emailAddress=a@b.com` currently used by OMV), but I guess the most "canonical" reference I have is the [`openssl` manual](https://www.openssl.org/docs/man1.1.1/man5/x509v3_config.html) itself:

>  **Subject Alternative Name**.
>
> The subject alternative name extension allows various literal values to be included in the configuration file. These include **email* (an email address) URI a uniform resource indicator, **DNS** (a DNS domain name), **RID** (a registered ID: OBJECT IDENTIFIER), **IP** (an IP address), **dirName** (a distinguished name) and otherName.
>
> The email option include a special ‘copy’ value. This will automatically include any email addresses contained in the certificate subject name in the extension. 
>
>  Examples:
>
> ```
> subjectAltName=email:copy,email:my@other.address,URI:http://my.url.here/
> subjectAltName=IP:192.168.7.1
> subjectAltName=IP:13::17
> subjectAltName=email:my@other.address,RID:1.2.3.4
> subjectAltName=otherName:1.2.3.4;UTF8:some other identifier
>
> subjectAltName=dirName:dir_sect
>
> [dir_sect]
> C=UK
> O=My Organization
> OU=My Unit
> CN=My Name
>```

MestreLion:
Actually the wording has improved in the latest manpages (I've used above the one from `1.1.1` as that's the one in OMV6)

https://www.openssl.org/docs/manmaster/man5/x509v3_config.html

Also tangent to the topic:
- https://www.openssl.org/docs/manmaster/man1/openssl-req.html

MestreLion:
Btw, `-new -batch -sha256` arguments are redundant in OMV's use-case:
- `-new` is implied by `-x509`
- `-batch` is not necessary when using `-subj` (which replaces the distinguished name, disabling prompts)
- `-sha256` is already the default in `openssl` 1.1.1 (and OMV6 uses `1.1.1w`)

votdev:
> Btw, `-new -batch -sha256` arguments are redundant in OMV's use-case:
> 
>     * `-new` is implied by `-x509`
> 
>     * `-batch` is not necessary when using `-subj` (which replaces the distinguished name, disabling prompts)
> 
>     * `-sha256` is already the default in `openssl` 1.1.1 (and OMV6 uses `1.1.1w`)

Thx. I've added them with reason because even if they are redundant, it shows better what is happening here without knowing the exact meaning of the arguments.

votdev:
> Note 2: Isn't it supposed to have a Comment field in the SSL certificate creation form? I can see it in the code (both in the HTML template and in the form processing code), and I also noticed a recent commit changing it from Text Input to Tag Input, but I don't see the field anywhere in the UI form (using latest OMV6):

grafik

The `Tags` field is at the bottom of the form.

votdev (approved):

votdev:
LGTM. Thx

Issue #1663, opened 2023-12-20
There needs to be a way to edit the hosts files


ElementalWarrior:
**Is your feature request related to a problem? Please describe.**
I can't edit the hosts file without editing salt files.

**Describe the solution you'd like**
A text box in the ui, or the hosts file is not fully replaced, just a block modified, sort of like how ansible handles it.

**Describe alternatives you've considered**
I worked with the salt files.

**Additional context**
Any save to etc/hosts makes scripting changes with things like ansible painful.

votdev:
Can you please give an example in which scenario it is necessary to modify this file manually?

ElementalWarrior:
I am running a dns server on omv and modifying hosts is what enables me to route hostnames like `wiki.home` to the omv server.

donhwy:
I have asked for this before. Use case I have a server that get letsencrypt cert. It is not in dns because only servers need to know about it. I use hosts file for this. OMV needs to know to get the new cert. Applies to other things too. I can use the ip in the scripts but names are nicer. Took a few times to figure out why things were broken. It is a basic debian feature.
Thanks

ryecoaaron:
The saltstack has already been changed to allow this with the creation of a simple salt file:
sudo vi /srv/salt/omv/deploy/hosts/20test.sls
```
configure_hosts_extras:
  file.append:
    - name: "/etc/hosts"
    - text: |
        #test
        192.168.1.1 test
````
sudo omv-salt deploy run hosts

donhwy:
Well I admit to not understanding the salt thing. Many more people know and understand the /etc/hosts file. Wouldn't it be better to have the salt understand and respect the /etc/hosts file? Maybe that is not possible?

ryecoaaron:
It is possible to have a section between comments that OMV controls like fstab but it would be even easier to have conflicting entries.  And then someone would want to tweak entries for localhost.  It wouldn't be hard to make a plugin but I have a feeling more people would break things than other users would benefit from being able to add entries.  I would either say OMV adds a boolean environment variable that disables managing /etc/hosts or people add the file I suggested since it is almost as easy as editing /etc/hosts itself.

votdev:
I've created a plugin for that. Also in the hope that this can be used as a template for other developers. There are of course risks involved in using the plugin, but the responsibility lies with the user.

PR #1664, opened 2023-12-20
A plugin to manage additional entries in /etc/hosts

votdev:
grafik

Fixes: https://github.com/openmediavault/openmediavault/issues/1663


ryecoaaron (approved):
works well and simple to use.

PR #1688, opened 2024-01-22
Use apticron to notify users about installable upgrades

votdev:


ryecoaaron:
Sorry I didn’t respond to this in time. Still out of the country and wifi has been spotty. Seems fine to me though.

PR #1709, opened 2024-02-11
Update omv-btrfs-scrub: only send mail if the scrub command found an error 

mael2:
Update omv-btrfs-scrub: only send mail if the scrub command found an error
try to reduce mails from OMV.
PIPESTATUS only works in bash.
Signed-off-by: Florian Schröck [mael2@mailbox.org](mailto:mael2@mailbox.org)

This is my first pull request ever.
Let me know, if there is something to improve :)


ryecoaaron:
In my opinion, this should be configurable not error-only.  Some users will want to receive an email on success.

mael2:
I didn't dive into the project this deep yet, but that would be great!

votdev:
> In my opinion, this should be configurable not error-only. Some users will want to receive an email on success.

I'm also in favour of sending emails when the scan is suceeded, not only if it fails.

votdev:
Please upgrade the PR and subscribe the DCO.

deb/openmediavault/usr/sbin/omv-btrfs-scrub
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env dash
+#!/bin/bash
```
votdev, line 1: Please use the path returned by `which`.
```suggestion
#!/usr/bin/bash
```

votdev (changes requested):

mael2:
I don't understand how to add the signoff to a single commit where it's missing. I think I made it worse. let's forget about it.

ryecoaaron:
@mael2 just use the --signoff flag when you commit a change.  If you want to sign an existing commit, just use git commit --amend --signoff

Issue #1730, opened 2024-03-22
Use sec to replace rsyslog


votdev:
Rsyslog is currently only used for looking into syslog and sending emails if there are log messages from `pamfaillock`.

- https://packages.debian.org/bookworm/sec
- https://github.com/simple-evcorr/sec
- https://simple-evcorr.github.io/man.html
- https://github.com/simple-evcorr/rulesets
- https://www.admin-magazine.com/Archive/2014/20/Simple-Event-Correlator
- https://www.usenix.org/system/files/login/articles/09_lang-online.pdf

votdev:
I have to reject this issue because rsyslog cannot be removed because it is used for syslog forwarding. This is very sad because sec seems to be a pretty cool feature.
Sending emails based on found log entries can also be done by rsyslog; which is done for pam_faillock messages.

PR #1732, opened 2024-03-22
Replaced rsyslog by sec (simple event correlator)

votdev:
Fixes: https://github.com/openmediavault/openmediavault/issues/1730

- [x] References issue


votdev:
Closing because rsyslog cannot be removed because it is used for syslog forwarding.

PR #1743, opened 2024-04-16
Homes Directory Fix for Active Directory users

yew2362:
Active directory fix. When logging using a AD you login as CORP\username so the folder is CORP/username and that errors out and %U will only set the username

When having Samba connected to an AD DC Server you are required to connect to the server as shown above (CORP\username) which means when you are logged in and you use %S it will show CORP\username and that isn't what you want to create as your home directory (Which would be useful). But that doesn't work when creating a folder. I'm not sure why since I even tried making the folder called HOME (or CORP in the example) and it would still give an error and it would only work when i made the all the folders manually. Home directory would be a great feature on Active Directory if it worked thank you!

Signed-off-by: yew2362

example error from samba (HOME = workgroup of ad dc):
[2024/04/13 18:20:23.813679,  0] ../../source3/smbd/smb2_service.c:772(make_connection_snum)
  make_connection_snum: canonicalize_connect_path failed for service m.brode, path /home/HOME/username


votdev:
Please check https://github.com/openmediavault/openmediavault/blob/master/CONTRIBUTING.md how to prepare a PR correctly.

votdev:
BTW, you can set the OMV_SAMBA_HOMES_VALIDUSERS environment variable according to https://docs.openmediavault.org/en/stable/various/advset.html#environmental-variables to customise the OMV defaults to your needs.

yew2362:
Thanks... I've tried doing the contributing correctly but I didn't do the signed off correctly.

PR #1755, opened 2024-05-02
Fix VirtualBox support

Forage:
The VirtualBox provider did not work as is.

- Made `patch.sh` libvirt only, avoids VirtualBox (re)boot issues
- Use integrated disk support, prevents `VERR_ALREADY_EXISTS` errors

As far as I can tell, the issue addressed with `patch.sh` only applied to libvirt anyway.
The way disk were added caused issues like https://github.com/hashicorp/vagrant/issues/8107. The only 'drawback' of the new way of adding the disks is that they are put in the default VirtualBox location instead of the `disks` folder with the Vagrantfile file.

### DIFF
--- a/vagrant/debian/README.md
+++ b/vagrant/debian/README.md
@@ -15,11 +15,6 @@ $ sudo apt-get install nfs-kernel-server
 $ sudo apt-get install vagrant
 ```
 
-4. Install vagrant-reload plugin (https://github.com/aidanns/vagrant-reload/tree/master)
-```
-$ vagrant plugin install vagrant-reload
-```
-
 # Using libvirt
 
 1. Install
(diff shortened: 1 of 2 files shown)


votdev:
Is the patching not necessary on a VirtualBox installation?

votdev:
The `signed-off-by` line is missing in the commit message. Please check https://github.com/openmediavault/openmediavault/blob/master/CONTRIBUTING.md how to contribute a PR.

Forage:
> Is the patching not necessary on a VirtualBox installation?

I have the following on a new VB setup, without the patch:
image

I can connect to the GUI without issues without the patch. These two devices is what you are after, wasn't it?

The IP address is the only additional thing I manually changed in the Vagrantfile, since the IP range did not correspond to the default range allowed by VB.

If I keep the patch, the machine will boot, but Vagrant is no longer able to cleanly finish its setup after the reload.

Forage:
> The `signed-off-by` line is missing in the commit message. Please check https://github.com/openmediavault/openmediavault/blob/master/CONTRIBUTING.md how to contribute a PR.

I couldn't fix the commit message while keeping the existing PR. New PR #1756.

ryecoaaron:
All you have to do is append the commit and force push.

PR #1771, opened 2024-05-31
Refactor the download.php RPC infrastructure

votdev:
... and make use of it in the RRD dashboard widget and page.

- [ ] References issue


ryecoaaron:
is this a breaking change if I am using contenttype outside of the new headers?   https://github.com/OpenMediaVault-Plugin-Developers/openmediavault-wireguard/blob/master/usr/share/openmediavault/engined/rpc/wireguard.inc#L428

votdev:
> is this a breaking change if I am using contenttype outside of the new headers? https://github.com/OpenMediaVault-Plugin-Developers/openmediavault-wireguard/blob/master/usr/share/openmediavault/engined/rpc/wireguard.inc#L428

No, content type is auto-detected and can be overridden by the headers field.

PR #1779, opened 2024-06-21
Show the drive power state in the drives table

ror3d:
Basically I'm adding the missing table column to the drives section that had the data gathering already implemented for it in `OMV\System\Storage\StorageDeviceSD`.

I added it to the drives section since it seems that just by going to the SMART section, the drives are spun up because the SMART calls wake them up.

Also, I wanted to mention that, according to some stackoverflow answers there's some hint that `hdparm` might sometimes also wake the drives, and I've seen it recommended to use `smartctl` instead, so maybe it will need to be changed at some point to avoid that. I looked into the `hdparm` issues and changelogs and there's no mention of this problem there, so either no one has reported it, it has been silently fixed, or is still there for some drives. Here's the references to this issue: https://serverfault.com/a/204235 https://serverfault.com/questions/275364/get-drive-power-state-without-waking-it-up

### DIFF
--- a/deb/openmediavault/usr/share/openmediavault/engined/rpc/diskmgmt.inc
+++ b/deb/openmediavault/usr/share/openmediavault/engined/rpc/diskmgmt.inc
@@ -84,6 +84,7 @@ class DiskMgmt extends \OMV\Rpc\ServiceAbstract {
 				"devicefile" => $sd->getPredictableDeviceFile(),
 				"devicelinks" => $sd->getDeviceFileSymlinks(),
 				"model" => $sd->getModel(),
+				"powerstate" => (($sd instanceof \OMV\System\Storage\StorageDeviceSD) ? $sd->getPowerModeState() : "unknown"),
 				"size" => $sd->getSize(),
 				"description" => $sd->getDescription(),
 				"vendor" => $sd->getVendor(),
--- a/deb/openmediavault/workbench/src/app/pages/storage/disks/disk-datatable-page.component.ts
+++ b/deb/openmediavault/workbench/src/app/pages/storage/disks/disk-datatable-page.component.ts
@@ -84,6 +84,33 @@ export class DiskDatatablePageComponent {
         flexGrow: 1,
         sortable: true,
         cellTemplateName: 'binaryUnit'
+      },
+      {
+        name: gettext('Power State'),
+        prop: 'powerstate',
+        flexGrow: 1,
+        sortable: true,
+        cellTemplateName: 'chip',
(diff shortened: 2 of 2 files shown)


votdev:
Hmmm, `hdparm` is not sed and installed by OMV anymore. So you've found code that is not working anymore. If possible, the `getPowerModeState` method must get the current power state from `smartctl`, or it has to be removed.

votdev:
When i read the [manual page of smartctl](https://www.smartmontools.org/browser/trunk/smartmontools/smartctl.8.in#n) i see that there is a problem that can't be solved.

Using `--nocheck=POWERMODE[,STATUS[,STATUS2]]` requires `-d` to do not spin-up the disk, but OMV does not support that, instead it fully relies on the auto-detection feature of smartmontools which will spin-up disks.

votdev:
@ror3d Would you please have a look at https://github.com/openmediavault/openmediavault/pull/1781?

votdev:
Closing in favor of https://github.com/openmediavault/openmediavault/pull/1781

ror3d:
> Using `--nocheck=POWERMODE[,STATUS[,STATUS2]]` requires `-d` to do not spin-up the disk

In my (limited) experience, I have never had to provide `-d` for it to not spin-up, but I suppose as it says in de9fb52 it depends on the drives being in the database for it to always work. How frequently is that expected to be a problem? it might be that smartmontools caches that information itself already, I've been looking at the code for it but it's not straightforward to find out.

Just so I understand, I expect the reason for not supporting `-d` is because it would require keeping a local database with the information gathered previously, right?

votdev:
> How frequently is that expected to be a problem?

I can't say anything about this.

> Just so I understand, I expect the reason for not supporting -d is because it would require keeping a local database with the information gathered previously, right?

In previous versions, I tried to identify the correct option for individual devices using heuristics and other information. This always took a lot of time as I did not have access to the hardware. In the end, the time required was too great and the result was error-prone.

The reason why I do not let the user configure the option in the UI is that OMV should be easy to use and the UI should not be overloaded with options that only very few users can do anything with. In addition, this would not help because `smartctl` is called in so many places where it would not even be possible to apply the `-d` option that the user would have configured in the UI.

It is therefore best to leave the identification of the devices to the smartmontools themselves; it is not without reason that they have a database intended for this purpose. This database is automatically updated by OMV once a week.

In addition, smartmontools offers the option of manually expanding this database; users can therefore enter their devices there themselves until they are integrated into the official database. Provided that these users contribute to this database with a code contribution.

Issue #1619, opened 2023-10-27
SMART support for SSD


matthabjan:
**Is your feature request related to a problem? Please describe.**
I have a similar issue as described here: #1467 
I have a Samsung SSD drive connected to Raspberry Pi4 via USB, where I can read smart data with the following command, after updating to smartmontools 7.3:

```smartctl -a /dev/sda -d sntasmedia``` 

But OMV throws an scsi error in the SMART tab.

smartctl 7.3 2022-02-28 r5338 [aarch64-linux-6.1.21-v8+] (local build)
Copyright (C) 2002-22, Bruce Allen, Christian Franke, www.smartmontools.org

Read Device Identity failed: scsi error unsupported scsi opcode

If this is a USB connected device, look at the various --device=TYPE variants
A mandatory SMART command failed: exiting. To continue, add one or more '-T permissive' options.

**Describe the solution you'd like**
Use SMART functions in OMV for this drive. I thought this issue was resolved in the newest OM releases. I am on OMV 6.9.4 now.

**Describe alternatives you've considered**
I have tried to run the drive in UAS and USB-storage mode, neither of which changes the issue.

votdev:
Same procedure as [here](https://github.com/openmediavault/openmediavault/issues/1588) .

udevadm info --query=all --name=/dev/sda

matthabjan:
Thanks for taking a look at this!

My output for ```udevadm info --query=all --name=/dev/sdb``` is

```P: /devices/platform/scb/fd500000.pcie/pci0000:00/0000:00:00.0/0000:01:00.0/usb2/2-1/2-1.3/2-1.3:1.0/host1/target1:0:0/1:0:0:0/block/sdb
N: sdb
L: 0
S: disk/by-id/usb-Samsung_PSSD_T7_S6XJNS0W413844M-0:0
S: disk/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.3:1.0-scsi-0:0:0:0
E: DEVPATH=/devices/platform/scb/fd500000.pcie/pci0000:00/0000:00:00.0/0000:01:00.0/usb2/2-1/2-1.3/2-1.3:1.0/host1/target1:0:0/1:0:0:0/block/sdb
E: DEVNAME=/dev/sdb
E: DEVTYPE=disk
E: DISKSEQ=27
E: MAJOR=8
E: MINOR=16
E: SUBSYSTEM=block
E: USEC_INITIALIZED=11895924
[...]

PR #1781, opened 2024-06-21
Refactor disk power mode detection

votdev:
`hdparm` has been removed previously, so the current implementation does not work anymore. `smartctl` will replace `hdparm` to detect the power mode of a disk.

- Add `Power Mode` column to `Storage | Disks` page.
- Add `Temperature` column to `Storage | Disks` page (since the SMART values have already been read out anyway and it therefore costs nothing).

Please note that not all power modes may be displayed correctly as they are not specified in `map`; this may affect NVMe devices in particular. These power modes must then be added subsequently.

Superseeds: https://github.com/openmediavault/openmediavault/pull/1779

- [ ] References issue

### DIFF
--- a/deb/openmediavault/debian/changelog
+++ b/deb/openmediavault/debian/changelog
@@ -1,6 +1,7 @@
 openmediavault (7.3.0-1) stable; urgency=low
 
   * Refactor the RPC parameter handling.
+  * Display `Power Mode` of disks in the `Storage | Disks` page.
   * Issue #1774: Improve foundation for plugins.
   * Refactor UI to get rid of `::ng-deep`, which was an obstacle to
     migrating to Angular >= 15.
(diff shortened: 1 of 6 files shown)


ror3d:
Oh this looks much better, it's been 15 years since last I used php so looking at your changes I've seen a bunch of issues in my implementation attempt. Thanks you so much for taking care of it!

votdev:
It seems the `--nocheck` option is causing troubles with some devices. See https://forum.openmediavault.org/index.php?thread/53058-upgrade-to-7-3-0-1-breaks-storage-disks-display/#post394140

ror3d:
Can the error/exception be caught if it complains about that and report "unknown" for those devices?

votdev:
> Can the error/exception be caught if it complains about that and report "unknown" for those devices?

This is possible, but i suggest to add the device to the custom smartmontools database and contribute it to the upstream smartmontools project.

- https://github.com/openmediavault/openmediavault/issues/1619#issuecomment-1788959178
- https://docs.openmediavault.org/en/stable/faq.html
- https://forum.openmediavault.org/index.php?thread/50161-after-update-from-6-9-6-1-to-2-smart-stopped-working-w-error/
- https://forum.openmediavault.org/index.php?thread/51543-smart-statuses-are-not-working-on-omv7-debian-12-raspi4/&postID=382936#post382936

ror3d:
Ah yeah to fix it obviously it will depend on the drive being added to smartmontools, I just meant it so that the interface doesn't just break for all the drives because of one not being in the database, as they reported in https://forum.openmediavault.org/index.php?thread/53058-upgrade-to-7-3-0-1-breaks-storage-disks-display/

Issue #1747, opened 2024-04-22
Basic auth


albinmedoc:
**Is your feature request related to a problem? Please describe.**
I'm using authentik for MFA, I'm frustrated that I need to enter OMV login details every time. 

**Describe the solution you'd like**
Enable Basic authentication. This would NOT lead to a more insecure or secure application. 

By allowing Basic Auth it would allow authentik to pass the details for a seemless login. And that would result in a more secure application, because I can enable MFA, TOTP, etc. 

**Describe alternatives you've considered**
Develop some other type of authentication like OIDC or SAML.

votdev:
What do you expect from this feature request. Do you want to get the RPC `Session:login` to be called like 
```
$ curl http://omv7box.local/rpc.php -u "admin:admin" --data '{"service":"session","method":"login"}'
```
or do you want to have a dedicated `login.php` script that handles the login?
```
$ curl http://omv7box.local/login.php -u "admin:admin"
```

votdev:
> What do you expect from this feature request? Do you want to get the RPC `Session:login` to be called like
> 
> ```
> $ curl http://omv7box.local/rpc.php -u "admin:admin" --data '{"service":"session","method":"login"}'
> ```
> 
> or do you want to have a dedicated `login.php` script that handles the login?
> 
> ```
> $ curl http://omv7box.local/login.php -u "admin:admin"
> ```

This would look like this:
```
	final public function login($params, $context) {
		// Check if `Basic` authentication is used.
		if (array_keys_exists(["PHP_AUTH_USER", "PHP_AUTH_PW"], $_SERVER)) {
			$params['username'] = $_SERVER['PHP_AUTH_USER'];
			$params['password'] = $_SERVER['PHP_AUTH_PW'];
		}
```

albinmedoc:
I'm not completely familiar with the code base for OMV or how everything works. 

But based on the two options you suggested, option 2 should work. A specific path (e.g /login.php) where username and password are sent as Basic Auth. 

Option 2 is excluded as Proxy Auth cannot send any body/data.

votdev:
@albinmedoc Could you please have a look on https://github.com/openmediavault/openmediavault/pull/1785 and test it?

albinmedoc:
This is not really what I wanted. 
After login, the user should be redirected to the dashboard. 

This is not for making API request. 

By allowing basic auth, another IDP like Authentik or Authenlia could be used to log in to OMV.

votdev:
[...]

PR #1785, opened 2024-06-25
Add support for basic auth

votdev:
- Add new login.php endpoint

To trigger a login, simply call `curl http://omv7box.local/login.php -u "admin:admin"` from a command line or an equal call via Python request/whatever.

On successful log in the following response will be returned:
```
{"response":{"authenticated":true,"username":"admin","permissions":"role":"admin"},"sessionid":"5o0lbvnsbr6r1dt6ql4hm5v3be"},"error":null}
```
The session ID is only returned at the first log in attempt of a user. It is also submitted via
```
Set-Cookie OPENMEDIAVAULT-SESSIONID=5o0lbvnsbr6r1dt6ql4hm5v3be; path=/; HttpOnly; SameSite=Strict
```
header.

The session ID can be sent via the `X-OPENMEDIAVAULT-SESSIONID` header or cookie on all following RPC requests.

Fixes: https://github.com/openmediavault/openmediavault/issues/1747

- [ ] References issue

### DIFF
--- a/deb/openmediavault/usr/share/php/openmediavault/rpc/proxy/json.inc
+++ b/deb/openmediavault/usr/share/php/openmediavault/rpc/proxy/json.inc
@@ -101,10 +101,23 @@ class Json {
 	}
 
 	/**
+	 * Set the RPC parameters.
+	 * Note, this is only for internal use.
+	 */
+	public function setParams($params) {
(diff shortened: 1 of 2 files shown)


wtrocki:
Nice this opens an really easy way to script shutdown of the server.
Thank you so much

PR #1790, opened 2024-07-01
Leave "ipv6-privacy" options in netplan to system defaults

graealex:
### Leave "ipv6-privacy" options in netplan to system defaults (either /etc/sysctl.conf or /etc/systemd/network/...) instead of manually overriding without any option for users to change.

OMV lacks an option to enable/disable IPv6 privacy extensions. Currently it overrides privacy extensions for every interface to be set enabled. Since there is no option, it should leave it at the default values set by either /etc/sysctl.conf or configuration files in /etc/systemd/network/ instead of trying to enforce them to be enabled.

### DIFF
--- a/deb/openmediavault/srv/salt/omv/deploy/systemd-networkd/netplan/files/bond.j2
+++ b/deb/openmediavault/srv/salt/omv/deploy/systemd-networkd/netplan/files/bond.j2
@@ -51,9 +51,6 @@ network:
 {%- if interface.method6 == "static" %}
       accept-ra: false
 {%- endif %}
-{%- if interface.method6 in ["static", "auto"] %}
-      ipv6-privacy: true
-{%- endif %}
 {%- if interface.method == "manual" and interface.method6 == "manual" %}
       addresses: []
 {%- endif %}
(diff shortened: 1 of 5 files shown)


graealex:
"There is one commit incorrectly signed off. This means that the author of this commit failed to include a Signed-off-by line in the commit message."

Now what did I do wrong this time?

votdev:
Are there good reasons to do not enable IPv6 privacy extensions?

graealex:
TL;DR - it makes servers (the obvious role of a host running OMV) change its own IPv6 address frequently. It also adds unnecessary "address noise" by doubling the number of IPv6 addresses a host gets assigned (usually LLA and GUA, optionally ULA - so 4/6 addresses instead of 2/3 addresses).

Longer explanation: when a client accesses a resource, for example a website, it would do so with the address acquired through SLAAC. Even with the ISP regularly assigning a new network _prefix_, the _host_ part would be produced from the MAC address through EUI-64 and remain static. Privacy extensions avoid this problem by regularly randomizing the address. Fine for a client, not very useful for a server that needs to be resolved through DNS.

My personal opinions:

1. Currently the configuration process of OMV overwrites system-wide defaults defined by sysctl.conf AND systemd-network, two places that could and should be able to define whether a user wants PE enabled or not - it should not override these defaults, the PR fixes this issue by leaving enabling/disabling PE to the system-defined defaults.
2. OMV should present a configuration option in its web UI to configure PE for every interface. It would be a simple addition.

PR #1792, opened 2024-07-05
Add StorageDeviceInterface::isHotPluggable() method

votdev:
Related to: https://forum.openmediavault.org/index.php?thread/53164-usb-backup-sata-drives-are-listed-but-sas-are-not/

- [ ] References issue


votdev:
@ryecoaaron The changed interface might affect your plugins; but only if you do not derive from the `\OMV\System\Storage\StorageDevice` class.

ryecoaaron (approved):
Looks ok to me.  The zfs and luks plugins both use the StorageDevice class.  I will look at the changes I need to make.

Issue #1799, opened 2024-07-27
openmediavault-update-smart-drivedb is executed weekly with bash -l


gromeck:
`/etc/cron.weekly/openmediavault-update-smart-drivedb` is the only cronjob script which uses `#!/usr/bin/bash -l` for its execution.
As the `-l` is used, the output of the script always contains messages which normally only login-shells would see (e.g. a dynamic greeting).
The effect is that the weekly execution of `openmediavault-update-smart-drivedb` will always send a mail.

**To Reproduce**

1. login to omv instance as root via ssh
2. execute
     `echo -e '#!/bin/bash\necho *** Welcome to $( hostname --short ) ***' >/etc/profile.d/welcome.sh`
4. re-login to omv instance as root via ssh
5. observe that the welcome message ("`*** Welcome to ... ***`") appears
6. wait until saturday morning
7. check your mailbox
8. observe that a mail with the welcome message was received

**Expected behavior**

No mail should be received.

**Screenshots**

n/a

**Reference to Forum**

n/a

**openmediavault Server:**

- OS version: Linux omv 6.1.0-23-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.99-1 (2024-07-15) x86_64 GNU/Linux
- openmediavault version: 7.4.3-1

**Client:**

n/a

**Additional context**

n/a

**Proposed Fix**

Please change the first line in `/etc/cron.weekly/openmediavault-update-smart-drivedb` from

    #!/usr/bin/bash -l

to

    #!/usr/bin/env dash

as it is done in all the other openmediavault cron scripts.

Thanks a lot for openmediavault,
gromeck

votdev:
This was intentionally done by https://github.com/openmediavault/openmediavault/commit/617ff78654a790feff3c03fd114dc0403c930ba4. 

Can you confirm that it works with a `#!/usr/bin/env dash` shebang?

gromeck:
Hmm, understood. Quite ugly that this is only the case for `openmediavault-update-smart-drivedb`.

I made the proposed change on my system (`#!/usr/bin/env dash` shebang) and moved the cron job testwise from weekly to hourly as I didn't wan't to wait until next saturday ;-)  I can see it's execution in the syslog, with no mail generated -- as expected.

To test the proxy stuff, I configured also a proxy host in openmediavault and added an `export` command to the cronjob script. The cron mail shows hat the environment has `http_proxy` only set, if the shebang uses the `-l` option.

[...]

PR #1800, opened 2024-07-27
Do not use a login shell

votdev:
Fixes: https://github.com/openmediavault/openmediavault/issues/1799

- [x] References issue


gromeck (changes requested):
As the proxy settings are only needed for the `update-smart-drive-db`, sourcing `/etc/profile` should be done only if it is really called (after the check if this smart drive test is enabled).

gromeck (changes requested):
Sourcing `/etc/profile` is in fact the same as using `-l` in shebang. Far that reason the change would not solve #1799.
To stay in the openmediavaul context, it would make sense to source `/etc/profile.d/*openmediavault*`

Issue #1816, opened 2024-08-20
[FR] Fix duplicate serial ID of an Insignia 2-Bay HDD docking station


yuchenshi:
**Is your feature request related to a problem? Please describe.**

I have this [hard drive dock](https://devicehunt.com/view/type/usb/vendor/0578/device/0578) from Insignia and it wrongly reports the same serial ID for the two hard drives. This leads to some issues in OMV too, such as two duplicate IDs showing up in the Disks and S.M.A.R.T. pages in the workbench.

Configuration created for one disk will automatically be applied to the other disk because OMV sees both as having the same serial number. This can also be seen in the XML config file, which will contain stanzas created by the workbench with the wrong ID.

I'd like to see the correct serials instead and individually control the two disks.

**Describe the solution you'd like**

I believe a udev rule similar to #1532 will fix this issue. I'd be happy to contribute a PR. I have a fix locally and I have verified that the fixed serial numbers are the real serial numbers as printed on the HDDs.

**Describe alternatives you've considered**

- I can configure a separate udev file on my server, but it doesn't fix the problem for anyone having the same device
- I can probably file an issue upstream (maybe to udev, idk) but I'm not sure if they'd be interested and/or their release cadence
    * In the meantime, fixing it in OMV seems like a good idea given the precedence of #1532 and other udev rules in that file
    * If this ever get fixed upstream, I'd be happy to create another PR to clean up the obsolete code in OMV
- I can ignore the issue and just carry on but I'll run the risk that something may break in the future due to the duplicate serial ID or the device file / symlinks
    * Maybe one day I'd try to format one drive and somehow the other drive also got wiped

**Additional context**

I have [posted on the forum](https://forum.openmediavault.org/index.php?thread/47388-qnap-tr-004-usb-raid-enclosure-duplicate-serial-number-fix/&postID=398618#post398618) first and was asked to create the issue here as a feature request.

PR #1817, opened 2024-08-20
 Add UDEV rule to fix the duplicate serial number issue on Insignia NS-PCHDEDS19 devices

yuchenshi:
Reference: https://forum.openmediavault.org/index.php?thread/47388-qnap-tr-004-usb-raid-enclosure-duplicate-serial-number-fix/&postID=398618#post398618
Reference: https://github.com/openmediavault/openmediavault/issues/1816

- [x] References issue


votdev (approved):
Thx

Issue #1814, opened 2024-08-18
Custom page route causing 404


ror3d:
**Describe the bug**

When creating a plugin and defining a route in a route.d/...yaml, if the route is too deep, it fails to be found and leads to 404 page.

**To Reproduce**

(I am putting the code for this with the :uuid parameter because this is how I have tested this to happen, because it is how I initially encountered the problem, but I believe without it a similar problem would occur)

1. Create two new test components in `/usr/share/openmediavault/workbench/component.d`:
    `omv-services-test-link-page.yaml`
    ```
    version: "1.0"
    type: component
    data:
      name: omv-services-test-link-page
      type: textPage
      config:
        title: "Test"
        buttons:
          - text: "Go To Test Page"
            url: "/services/test/foo/bar/baz/203d18f5-130b-430c-beb4-ca2efa8f406e"
    ```
    and the one that will 404 (using some form for some content)
    `omv-services-test-text-page.yaml`:
    ```
    version: "1.0"
    type: component
    data:
      name: omv-services-test-text-page
      type: textPage
      config:
        title: Test
    ```
3.  Create routes to the test components in `/usr/share/openmediavault/workbench/route.d`:
    `services.test.yaml`:
    ```
    version: "1.0"
    type: route
    data:
      url: "/services/test"
      title: _("Test")
      component: omv-services-test-link-page
    ```
    `services.test.text.yaml`:
    ```
    version: "1.0"
    type: route
    data:
      url: "/services/test/foo/bar/baz/:uuid"
      title: _("Test")
      component: omv-services-test-text-page
    ```
5. Compile the new components with `omv-mkworkbench all`
6. Navigate to `<server>/#/services/test`
7. Click the button on the page to Go To Test Page
8. Notice the 404 page appears

**Expected behavior**

The text page should load normally

**openmediavault Server:**

- OS version: Linux omv-dev 6.1.0-23-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.99-1 (2024-07-15) x86_64 GNU/Linux
- openmediavault version: 7.4.5

**Client:**

- Device: desktop
- OS: windows
- Browser: Vivaldi
- Version 6.2

**Additional context**

I have traced the routing that happens in the webapp (this was quite painful) and determined that the issue is in the `RouteConfigService`.

[...]

PR #1818, opened 2024-08-20
Issue #1814: Improve Angular route injection

votdev:
Fixes: https://github.com/openmediavault/openmediavault/issues/1814

- [ ] References issue


votdev:
I'm not sure if this fix contradicts the original idea that plugin developers have to configure ALL routes they use.  This ALWAYS ensures that the routing works as intended. This fix can cover bugs.

PR #1839, opened 2024-10-25
Update the DNS setting to be a mandatory configuration ...

votdev:
... for both Ethernet and WiFi network interfaces when a gateway is set.

There are simply too many posts in the forum that are due to users not configuring the DNS server when configuring static IP.

- [ ] References issue


ryecoaaron (approved):
I think this is a good idea.

ryecoaaron:
I think you should force dns if there is no gateway set.

PR #1842, opened 2024-10-30
Add code editor page

votdev:
Add a read-only code editor page. The languages `json`, `python`, `shell`, `xml` and `yaml` are supported. Syntax highlightning and code folding is available for those languages.

image

- [ ] References issue


ryecoaaron:
Sorry for the late comment but this seems fine to me.

votdev:
> Sorry for the late comment but this seems fine to me.

No problem, it was more a notification for you that there is such a new feature.

Issue #1694, opened 2024-01-28
PWA Manifest.json


iecxiv:
Hey. you could add the manifest file.json to the workbench so that it can be installed on android as a pwa

**Describe the solution you'd like**
Include manifest.json 

Thanks in advance

votdev:
Sounds good, but this feature has to be contributed by the community because i do not have any Android device for testing.

iecxiv:
where is the web directory located?

votdev:
> where is the web directory located?

Check https://github.com/openmediavault/openmediavault/tree/master/deb/openmediavault/workbench

iecxiv:
i mean when it is installed on the system, thanks

votdev:
> i mean when it is installed on the system, thanks

You can find it under `/var/www/openmediavault`. But for what reason do you want to know that?

iecxiv:
to try to implement the manifest.json on my own. i have already tried but it does not work

votdev:
![Image](https://github.com/user-attachments/assets/ba3c1fa4-cecf-448a-bd3a-e094fc184852)

![Image](https://github.com/user-attachments/assets/64020d98-dd57-4659-bd31-3d6265d42383)

Kirkland-Pickles:
what ever happened with this? i see no pwa option currently

votdev:
> what ever happened with this? i see no pwa option currently

To be honest, I don't see anything positive about this feature but only negative side effects of the implementation that affect all users who don't really want to know about this feature. In my opinion, PWA is implemented in Angular in such a way that it only leads to problems, for example, there will be a lot of problems if the code of the applications changes. Since it is too risky to degrade the entire UI here, I will refrain from implementing the feature in the near future.

PR #1846, opened 2024-11-11
Initial commit PWA manifest #1694

a-st:


votdev:
Could you please upgrade the draft PR to Angular 15. OMV has been upgraded this week. Note, Angular Material still stays as 14 atm.

a-st:
Guess it's working as intended
image
image
image

a-st:
@votdev Is anything else missing in order to get this PR reviewed?

votdev:
> @votdev Is anything else missing in order to get this PR reviewed?

I'll do a review asap. But don't expect it to be merged this year. This change is too dangerous to be included into 7.x atm. Maybe in January after the christmas vacation time where OMV is installed by many new users.

deb/openmediavault/workbench/src/manifest.webmanifest
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "openmediavault",
+  "short_name": "openmediavault",
+  "theme_color": "#1976d2",
```
votdev, line 4:
```suggestion
  "theme_color": "#5dacdf",
```

deb/openmediavault/workbench/src/index.html
```diff
@@ -37,6 +38,8 @@
       animation: toggle-opacity 1s step-start infinite;
     }
   </style>
+  <link rel="manifest" href="manifest.webmanifest">
+  <meta name="theme-color" content="#1976d2">
```
votdev, line 42: The [corporate color](https://github.com/openmediavault/openmediavault/blob/d97586361c7fb9c0d830d2e281f57dc280fdb9ba/deb/openmediavault/workbench/src/scss/defaults/_colors.scss#L23) is #5dacdf.

```suggestion
  <meta name="theme-color" content="#5dacdf">
```

votdev (changes requested):

deb/openmediavault/workbench/src/manifest.webmanifest
a-st: Thanks a lot for the suggested changes. The color has been updated accordingly 👍

deb/openmediavault/workbench/src/index.html
a-st: Thanks a lot for the suggested changes. The color has been updated accordingly 👍

votdev:
Did some improvements.

Nevertheless, even without my changes, the service worker is not enabled in production mode. Everything works in development mode, but when i build the package using `fakeroot debian/rules clean binary` and install the package, then FireFox does not detect a runninmg service worker in the devtools apps tab. Also Chrome does not show me the icon to install the app.

votdev:
What happens when the app is updated. I did a quick search and it seems the service worker did not reload the app in that case. If a plugin is installed or the core package did some changes to the UI, how is the cached app reloaded?

votdev:
PWA does not work in HTTP mode and even with self-signed certificates. How to tell users to get PWA running in that case. I on't think most users have valid SSL certificates or know how to import self-signed certs into their browsers to prevent the browser warning which is the reason why the service worker is not registered.

To me this feature looks nice, but can't be handled by most users.

votdev (changes requested):
Tried to implement the SwUpdate service to reload the page, but i didn't get it running. The app data is not updated which means that pages are not updated if, for example, a new input field has been added or help texts have been changed. This makes this feature unsuitable as the UI changes with almost every package update.

PR #1855, opened 2024-11-19
Add `OMV_FSTAB_MNTOPS_REISERFS` and `OMV_FSTAB_MNTOPS_REISER4` env variables

votdev:
Use mount options that fit for most users. Power users can use the openmediavault-mounteditor (https://github.com/OpenMediaVault-Plugin-Developers/openmediavault-mounteditor) to adapt them to their needs.

Note, the filesystems can be mounted only when the `reiserfsprogs` and `reiser4progs` packages are installed. These package will not be installed by default.

[Bildschirmaufzeichnung vom 2024-11-19 13-38-19.webm](https://github.com/user-attachments/assets/52f6dc10-f26d-4014-88fe-d4ee140b5280)

- [ ] References issue

### DIFF
--- a/deb/openmediavault/usr/share/php/openmediavault/globals.inc
+++ b/deb/openmediavault/usr/share/php/openmediavault/globals.inc
@@ -90,6 +90,8 @@ define("OMV_NETWORK_INTERFACE_TYPE_ALL", 0xFF);
 \OMV\Environment::set("OMV_FSTAB_MNTOPS_UDF", "ro");
 \OMV\Environment::set("OMV_FSTAB_MNTOPS_BIND", "nofail");
 \OMV\Environment::set("OMV_FSTAB_MNTOPS_F2FS", "defaults,nofail,user_xattr,usrjquota=aquota.user,grpjquota=aquota.group,jqfmt=vfsv1"); // https://www.kernel.org/doc/Documentation/filesystems/f2fs.txt
+\OMV\Environment::set("OMV_FSTAB_MNTOPS_REISERFS", "defaults,nofail,user_xattr,acl");
+\OMV\Environment::set("OMV_FSTAB_MNTOPS_REISER4", "defaults,nofail,txmod=journal");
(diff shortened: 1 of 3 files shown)


deb/openmediavault/usr/share/php/openmediavault/globals.inc
```diff
@@ -90,6 +90,8 @@ define("OMV_NETWORK_INTERFACE_TYPE_ALL", 0xFF);
 \OMV\Environment::set("OMV_FSTAB_MNTOPS_UDF", "ro");
 \OMV\Environment::set("OMV_FSTAB_MNTOPS_BIND", "nofail");
 \OMV\Environment::set("OMV_FSTAB_MNTOPS_F2FS", "defaults,nofail,user_xattr,usrjquota=aquota.user,grpjquota=aquota.group,jqfmt=vfsv1"); // https://www.kernel.org/doc/Documentation/filesystems/f2fs.txt
+\OMV\Environment::set("OMV_FSTAB_MNTOPS_REISERFS", "defaults,nofail,user_xattr,acl");
+\OMV\Environment::set("OMV_FSTAB_MNTOPS_REISER4", "defaults,nofail,txmod=journal");
```
mckaygerhard, line 94: `txmod=journal` is only recommended for spinning disk.. hybrid is best for any one.. , i assume "nofail" comes from kernel api mount interface..

votdev: I think I have already mentioned several times how OMV approaches and implements certain things. OMV uses default values that probably suit 99% of users. As for the mount options, there is a plugin that allows the user to customize them.

Nevertheless, I have solved the problem by adding the mount option `txmod=wa` for non-rotating devices and not setting the option `txmod` otherwise so that the default value is taken.

Fixed with https://github.com/openmediavault/openmediavault/commit/477cbd24b448673f1020a17f7c562fd9754991de.

mckaygerhard: Thanks for support.. you are great!

PR #1868, opened 2024-12-06
Correct and clarify usermgmt.inc docs for disallowusermod setting

mordyovits:
The description of the disallowusermod boolean's effect was accidentally reversed. Also clarified that the setting's effect is only enforced within the OMV web interface.

- [ ] References issue


votdev (approved):

votdev:
Thx

PR #1881, opened 2025-01-01
add In Use column for logical volumes

ryecoaaron:
Add In Use column for logical volumes

Add In Use column on LV datatable.  Hidden by default.
Determine if that logical volume is in use in a mount point or tgt target and disable LV reduce if mounted or part of a tgt target.

Fixes: https://forum.openmediavault.org/index.php?thread/55098-lvm-resize-volume-will-cause-filesystem-corruption/

Signed-off-by: Aaron Murray [plugins@omv-extras.org]

### DIFF
--- a/deb/openmediavault-lvm2/usr/share/openmediavault/engined/rpc/logicalvolumemgmt.inc
+++ b/deb/openmediavault-lvm2/usr/share/openmediavault/engined/rpc/logicalvolumemgmt.inc
@@ -606,6 +606,9 @@ class OMVRpcServiceLogicalVolumeMgmt extends \OMV\Rpc\ServiceAbstract {
 			$lv = new \OMV\System\Storage\Lvm\LogicalVolume($lvv);
 			if (FALSE === $lv->exists())
 				continue;
+			$used = false;
+			if ($this->checkForMountedFilesystem($lv->getPreferredDeviceFile()))
+				$used = true;
 			$result[] = [
 				"devicefile" => $lv->getPreferredDeviceFile(),
 				"uuid" => $lv->getUuid(),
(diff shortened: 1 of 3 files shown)


votdev:
I'm not a fan of this enhancement because:

- It accesses data from another plugin.
- Other plugins such as MD (Multiple Devices) or BCache do not indicate whether a device is used by a file system.
- This will break UX because things are implement different on several plugins.

ryecoaaron:
It does access data from another plugin because it is impossible to know if a logical volume is being used for an tgt target without looking in the config file (or something involving another plugin).  A tgt target is also where an lvreduce is most dangerous.  The LVM plugin will reduce an LV deleting data.

The In Use column could be removed from the datatable then. It was really just to disable the reduce button.  That is why the column is hidden by default.

I'm not sure how it breaks UX since no other core OMV plugins have a reduce function.

Closing since I don't have an alternative for the first point.

votdev:
Not being a fan does not mean it has to be closed immediately.

Are there valid reasons for this function from which a user can benefit?

ryecoaaron:
> Are there valid reasons for this function from which a user can benefit?

Other than preventing a user from reducing an LV and losing data?  I don't think so.

votdev:
I've [fixed an really old feature](https://github.com/openmediavault/openmediavault/pull/1882) that can be used for this PR.

The TGT plugin can register a hook for the `LogicalVolumeMgmt::enumerateLogicalVolumes` RPC and iterate over the result array and modify the `inuse` property.

deb/openmediavault-lvm2/usr/share/openmediavault/engined/rpc/logicalvolumemgmt.inc
```diff
@@ -614,12 +618,106 @@ class OMVRpcServiceLogicalVolumeMgmt extends \OMV\Rpc\ServiceAbstract {
 				"vgname" => $lv->getVGName(),
 				"attributes" => $lv->getAttributes(),
 				"vgattributes" => $lv->getVGAttributes(),
-				"isavailable" => $lv->IsMediaAvailable()
+				"isavailable" => $lv->IsMediaAvailable(),
+				"inuse" => $inuse
```
votdev, line 622: Wold it make sense to rename the property to `isinuse` or better use the internal `_used` property. This property is used in various places and the datatable should display it automatically afaik.

ryecoaaron:
I will try to figure out the hook system if I have time.

deb/openmediavault-lvm2/usr/share/openmediavault/engined/rpc/logicalvolumemgmt.inc
```diff
@@ -614,12 +617,74 @@ class OMVRpcServiceLogicalVolumeMgmt extends \OMV\Rpc\ServiceAbstract {
...
+		}
+		return false;
+	}
+
+        /**
+         * Get paths for logical volume
+         * @param string $path any logical volume path
+         * @return An array containing three logical volume paths.
+         */
+	private function getLvmPaths($path) {
```
votdev, line 658: I think this function can be replaced by the `getDeviceFileSymlinks` function. You can call it with `$lv->getDeviceFileSymlinks()`.

On my test system udevadm returns this:
```
DEVLINKS=/dev/vg0/lv0 /dev/disk/by-id/dm-uuid-LVM-x3YAcIjrdCGAMaFA6lfp1bEh2T3uKieqsBYoFUyzoJCp1c48CypereE9M2F9ofHO /dev/mapper/vg0-lv0 /dev/disk/by-id/dm-name-vg0-lv0
```

deb/openmediavault-lvm2/usr/share/openmediavault/engined/rpc/logicalvolumemgmt.inc
```diff
@@ -614,12 +617,74 @@ class OMVRpcServiceLogicalVolumeMgmt extends \OMV\Rpc\ServiceAbstract {
...
 		return $result;
 	}
 
+	/**
+	 * Checks /proc/mount for a mount point for a given LVM logical volume path.
+	 *
+	 * @param string $lvPath The full path of the LVM logical volume (can be /dev/dm-0, /dev/vgname/lvname, or /dev/mapper/vgname-lvname).
+	 * @return boolean true if a mount point is found
+	 */
+	private function checkForMountedFilesystem($lvPath) {
```
votdev, line 633: I think it is not necessary to parse the mounts file; a simply database query for `fsname` of `conf.system.filesystem.mountpoint` that uses all device files returned by `getDeviceFileSymlinks` is enough here. Do you agree?

votdev (changes requested):

PR #1882, opened 2025-01-03
Add ability to mutate the response of a specified RPC

votdev:
Example:
```
class MyMethodHooks extends \OMV\Rpc\ServiceAbstract {
	public function getName() {
		return "MyMethodHooks";
	}

	public function initialize() {
		$this->registerMutatingMethod("System", "getShells", "myGetShells");
	}

	public function myGetShells($params, $context, $result) {
		print_r($result);
		$result[] = "/usr/bin/foo";
		return $result;
	}
...
```

- [ ] References issue


ryecoaaron (approved):
This seems to satisfies the need for the tgt plugin to alter LV rpc.  I just need to figure out how to use it.

Issue #1796, opened 2024-07-19
[feature reqest] Create a user or installation setting for number of items per page.


DjoeC:
**Is your feature request related to a problem? Please describe.**
On a large screen it is frustrating to have a maximum of 25 visible items.  On a mobile device when more than 25 items are available it requires some scrolling, followed by paging - 2 different actions.  Being able to have "all" or some user-determined arbitrary number or choice from a number of presets, saved between sessions, would be a nice enhancement. 

**Describe the solution you'd like**
I would like a per user/installation setting of the (maximum) number of items per page.  Refer to the ["limit"](https://github.com/openmediavault/openmediavault/blob/e73c835a537091ea4518324779a718514b070bb2/deb/openmediavault/workbench/src/app/core/components/intuition/models/datatable-page-config.type.ts#L53) property.

**Describe alternatives you've considered**
Currently there is no alternative but the standard limit of 25 items per page other than a (local) code adjustment.

**Additional context**
According to post https://forum.openmediavault.org/index.php?thread/53333-possible-to-increase-number-of-items-per-page-in-services-compose-files-containe/&postID=396202#post396202 the enhancement should be possible.

I understand that with summer being here it will not have great priority BUT the enhancement would be greatly appreciated.

votdev:
The used [datatable component](https://github.com/siemens/ngx-datatable) does not support that out-of-the-box.

DjoeC:
@votdev Thanks for keeping this alive as I would still like to have the option.  And sorry for not being able to contribute code myself.

a-st:
@votdev I'm currently preparing a PR for this. Would something like that be acceptable?

![Image](https://github.com/user-attachments/assets/4c601508-31e0-4546-ba6c-6eac9514148d)

![Image](https://github.com/user-attachments/assets/02884c9e-239e-430d-80f4-5225ca35f05b)

DjoeC:
> [@votdev](https://github.com/votdev) I'm currently preparing a PR for this. Would something like that be acceptable?
> 
> ![Image](https://github.com/user-attachments/assets/4c601508-31e0-4546-ba6c-6eac9514148d)
> 
> ![Image](https://github.com/user-attachments/assets/02884c9e-239e-430d-80f4-5225ca35f05b)

Also an "all" (or "0"=all) option?

Thanks!

a-st:
@DjoeC Good call. The input now allows all values >= 0 with 0=all

DjoeC:
[...]

PR #1900, opened 2025-01-22
Feature #1796: Allow user to change the (maximum) number of items per page

a-st:
Allow user to change the (maximum) number of items per page. Value is stored in local storage and restored on login.

- [x] References issue


votdev:
Thank you for your contribution. But to be honest, I don't like the solution and implementation at all.
From a UX perspective, this dropdown belongs in the pagination bar.
The solution with the service that stores the settings in the browser session does not harmonize at all with the implementation of the user settings such as the columns to be displayed or the sorting order of the table which is stored in the backend.
I have high expectations for the OMV UI. That's why I haven't implemented this functionality myself yet, because I didn't like any of the solutions I came up with. Customizing the pagination bar requires a lot of effort.

a-st:
Thanks for your honest feedback. Guess I'll have to look into implementing it for the upstream project ngx-datatable

Issue #1918, opened 2025-02-15
Tokens not being replaced in RPC requests when using formPage + datatable + formDialogConfig


DerDreschner:
**Describe the bug**

I'm currently developing a new plugin for OpenMediaVault. During development, I experienced a possible bug regarding `formDialog` elements, which leads to tokens not being processed when doing RPC requests.

I want to create a `formPage` with a `datatable` inside. This data table uses a `formDialogConfig` for the add button. For the fields in the form dialog, I have to make parameterized RPC calls using objects from the `PageContext` variables (`_routeQueryParams` and `_routeParams` to be precise). But no matter what I do, the tokens are not being replaced. The raw value is being pushed to the RPC service without any change.

This is unexpected, as the token replacement for the `formDialog` works when using a `datatablePage` instead. But this isn't a possible workaround in my case.

**To Reproduce**

1. Create a new page with the following component:

```yaml
version: "1.0"
type: component
data:
  name: omv-services-bugreport-form-page
  type: formPage
  config:
    request:
        service: ShareMgmt
        get:
          method: enumerateSharedFolders
    fields:
      - type: datatable
        label: _("Bugreport")
        columns:
          - name: _("Column 1")
            prop: column1
            flexGrow: 1
            cellTemplateName: copyToClipboard
          - name: _("Column 2")
            prop: column2
            flexGrow: 1
            cellTemplateName: copyToClipboard
        actions:
          - template: add
            formDialogConfig:
              title: _("Add new lorem ipsum")
              fields:
                - type: select
                  name: bugField
                  label: _("This is a bug")
                  store:
                    proxy:
                      service: bugService
                      get:
                        method: getBugreportField
                        params:
                          value: "{{ _routeQueryParams.exampleValue }}"
```

2. Open the new page with the queryParameter `?exampleValue=test` added to the URI.
3. Try to add a new element to the data table.
4. Check the request against the RPC service. It should look like this:

```json
{
	"method": "getBugreportField",
	"params": {
		"value": "{{ _routeQueryParams.exampleValue }}"
	},
	"service": "bugService"
}
```

**Expected behavior**

The token of the parameter should have been replaced with the "test" value from the `exampleValue` query parameter. The request body would look like this then:

```json
{
	"method": "getBugreportField",
	"params": {
		"value": "test"
	},
	"service": "bugService"
}
```

**openmediavault Server**

[...]

PR #1923, opened 2025-02-18
Feature: Move `PageContext` to singleton service and format tokens in `prop` constrain fields

DerDreschner:
During #1918, I experienced cases where the `pageContext` isn't set. I tried to fix the issue myself, but it was pretty hard to have an overview of all the places where it's passed along (and where it should, but wasn't). Therefore, I decided to rewrite the page context and move it into a singleton service as well as implement the token format in `prop` constrain fields.

That's the second revision, as you can see from the commit history. First, I used a scoped service that had to be injected into each component. Otherwise, the `ActivatedRoute` wasn't correct (as the service was being created on application start). But I wasn't very happy with that approach. It was hard to identify all the places that needed to change in my testing, and creating a new instance with each component was a mess to debug quite quickly. Instead, I did some additional research and found a way to achieve the same result without the `ActivatedRoute` service. See this [discussion](https://github.com/angular/angular/issues/11023) as well as this [pull request](https://github.com/angular/angular/pull/40306).

This should be a non-breaking change and behave exactly like the old approach. Except it's now ensured the page context is always available. But as I made sure the page context is the only "source of truth," I did some little refactoring as well. I didn't find any more bugs in my testing, but please test yourself as well.

One additional note: [In this function](https://github.com/DerDreschner/openmediavault/blob/4d6989da4eb70b057ea9a70c90c807889a0f87b4/deb/openmediavault/workbench/src/app/core/components/intuition/datatable-page/datatable-page.component.ts#L150), I found that the fields of the selected element are being made available as `_selected[0].{{fields}}` as well as directly on root level (as `{{fields}}`). I didn't want to make any changes here, as I don't know if it's a desired behavior or not and what implications a change would have. But as I found it a bit odd to have it available both ways, I left a TODO note for future reference.

- [x] References issue


DerDreschner:
By the way: The reason I did this rewrite was the empty pageContext in `formDialog` fields. Beside the little change [here](https://github.com/openmediavault/openmediavault/pull/1923/files#diff-828094e0aab9db08d62784c15e669973bdb09a07b255f9ef6a1ee4fbe99ba81f) regarding the second case in [#1918](https://github.com/openmediavault/openmediavault/issues/1918#issuecomment-2666005521), which has nothing to do with the refactoring itself.

```yaml
version: "1.0"
type: component
data:
  name: omv-services-bugreport-form-page
  type: formPage
  config:
    request:
      service: ShareMgmt
      get:
        method: enumerateSharedFolders
        params:
          value: "{{ _routeQueryParams.exampleValue }}"
    fields:
      - type: datatable
        label: _("Bugreport")
        columns:
          - name: _("Column 1")
            prop: column1
            flexGrow: 1
            cellTemplateName: copyToClipboard
          - name: _("Column 2")
            prop: column2
            flexGrow: 1
            cellTemplateName: copyToClipboard
        actions:
          - template: add
            formDialogConfig:
              title: _("Add new lorem ipsum")
              fields:
                - type: select
                  name: bugField
                  label: _("This is a bug")
                  value: "{{ _routeQueryParams.exampleValue }}" <--- Value not set as pageContext is empty in those form and formFields. Want to move information from the "main form" to the formDialog.
```

I wasn't able to find where the injection chain for the `pageContext` is broken. As already said, the `pageContext` is a global context in my opinion. It depends on the route - which is a global state - and should be updated/handled when the route changes. That's what I do with my subscription to the `RoutesRecognized`. Additionally, it should be available to _every_ element on the page with the _same_ data, which is hard to overlook with the current situation with all the injections as visible with the above example. Maybe it's more clear if it would be called `routeContext`. Yes, the `_selected` object is an exception to that. But I would move that to an separate service / context if it's disrupting here honestly.

Therefore, I think my approach is valid and maintainable. Additionally, my new service has 100% test coverage, which is an improvement as well. I would be happy if you could take a look at the code again. If you still dislike it, just close the PR.

Thanks.

deb/openmediavault/workbench/src/app/shared/services/constraint.service.ts
```diff
@@ -102,8 +102,10 @@ export class ConstraintService {
       let result;
       if (_.isPlainObject(node)) {
         if (_.has(node, 'prop')) {
+          const containsToken =
```
votdev, line 105: Is there a unit test for that. I would like to have an example where it is necessary to tokenize the property field.

votdev: IMO tokeinzed strings should be formatted at an earlier stage, e.g. https://github.com/openmediavault/openmediavault/commit/ce5d7b88f3a6657798161ffa56500885bfab71da

deb/openmediavault/workbench/src/app/core/components/intuition/abstract-page-component.ts
```diff
@@ -49,10 +44,7 @@ export abstract class AbstractPageComponent<T> implements AfterViewInit, OnInit,
   @Output()
   readonly afterViewInitEvent = new EventEmitter();
 
-  readonly routeUrlSegments: string[];
-  readonly routeConfig: Route;
-  routeParams: Params = {};
-  routeQueryParams: Params = {};
+  public pageContext: PageContext = inject(PageContextService).getContext();
```
votdev, line 47: Please use old-style dependency injection via constructor. All code in this project is doing it this way, so there is no reason to mix old and new style. If new style is preferred, all code needs to be refactored.

votdev:
Thank you very much for the contribution. I took it as inspiration to implement the idea according to my ideas, especially because I want to separate the different contexts according to their tasks and concerns.

Superseded by https://github.com/openmediavault/openmediavault/pull/1928.

PR #1926, opened 2025-02-20
TaskSshDialog integrated with omv-composer

gekomod:
Adding Task Ssh Dialog to Omv-composer


gekomod:
My question is, what do you think about integration with omv-compose, the ability to open a window with the ssh console from the www /services/compose/files level, e.g. a button to open the ssh dialog

votdev:
What is the requirement for this feature? Who wants to use it and for what reason? Is this meantto be a SSH console in the webUI? For exactly that reason there is the openmediavault-wetty plugin? How do you want to configure the UI part when the user is running SSH on a different port than 22?

gekomod:
This is a display only code, and the server to handle the connection is in a different file and its connection to ssh looks like this:
"
  ssh.connect({
    host: 'localhost',
    username: 'omv_docker',
    password: 'PASSWORD'
  });
"

votdev:
All well and good, but how will the backend access via SSH? What privileges will the user used for this have? Is there privilege escalation via sudo? To be honest, I don't like the whole thing for security reasons.

You know, I looked at the whole thing years ago and didn't integrate it because I don't like the whole stack that's needed for it. that hasn't changed even now.

gekomod:
I prepared it for composer, but this way you turn on the console, log in to ssh and execute commands there, so such permissions as you give to the omv_docker user - such access will be given to the server, I did not see the need for integration in something else, in omv composer there is no access to ssh via www, I took the example from portainer which is a combine harvester and nobody says anything there that the user has too many permissions etc. because who and why would anyone else go there?

votdev:
If this will be a general purpose enhancement then other plugins can make use of it as well, e.g.the Kubernetes plugin.

> I did not see the need for integration in something else

I see it differently. If it is only for a specific plugin or use-case then it is out. I don't want to have special things in the core code base that only work for certain plugins. If it's not general purpose, then it doesn't go in. That's my philosophy and it's not going to change.

> I took the example from portainer which is a combine harvester and nobody says anything there that the user has too many permissions etc. because who and why would anyone else go there?

That is naive.

gekomod:
of course if kuber will be needed why not then the server will be added to the main file system as you can see from the code it is added to the datatable so the usage can be general, originally I did not think about the kuber plugin because I did not use it, but the usage of it can be general.

deb/openmediavault/workbench/src/app/shared/components/task-dialog-ssh/task-dialog-ssh.component.ts
```diff
@@ -0,0 +1,225 @@
...
+  public running = false;
+  private subscription: Subscription;
+
+  constructor(
+    private authSessionService: AuthSessionService,
+    @Inject(MAT_DIALOG_DATA) data: TaskDialogSshConfig
+  ) {
+    this.config = data;
+    this.sanitizeConfig();
+    this.socket = io('http://127.0.0.1:1122', { }); // Connect to server Socket.IO
```
votdev, line 78: How do you plan to make the configuration of the socket configurable so that it can be used by all possible plugins? Since plugins can only make the configuration declaratively, I would be interested in how you want to implement this.

gekomod: physically the address to the server should not change, it should connect via localhost, because of CORS, I would also leave the data for SSH unchanged, only the port could be taken from omv

Issue #1887, opened 2025-01-09
bcachefs


colttt:
**Is your feature request related to a problem? Please describe.**
Since bcachefs is in the Kernel it would be nice if you add some support for, like create disk arreays and get the status etc, like ZFS/btrfs

ryecoaaron:
While I think it sounds neat, bcachefs is still very experimental/changing and not available in the standard Debian 12 kernel (6.1).  In my opinion, it is too early to start developing for it.

votdev:
If it is in the official Debian kernel and the userland tools in the package repository, then it might make sense to think about it.

votdev:
Done with https://github.com/openmediavault/openmediavault/pull/1853.

liberodark:
+1 for bachefs

ApexArray:
Chiming in to say that I’d love to see bcachefs support in OMV eventually, but that it’s not viable on Debian for most users at this time due to the orphaned userspace tools.

I’m a heavy user of bcache (non-fs) though and was pleasantly surprised to see the recent plugin added to support it.

Subscribed to this issue for future updates.

jaccoh:
Me too.. I'm looking into caching and I'd love to use bcachefs. Right now I'll test with bcache.

PR #1930, opened 2025-03-09
Init Bcachefs support

liberodark:
First work for this issue : https://github.com/openmediavault/openmediavault/issues/1887

This PR is made by @Nopraz & @liberodark 

- [x] References issue

image


votdev:
Why should this code be merged if the corresponding tools are not available in Debian Bookworm and Trixie?

No normal user will ever be able to make use of it because the userland apps ([bcachefs-tools package](https://packages.debian.org/sid/bcachefs-tools)) are only available in Debian Sid, which OMV does not make use of.

liberodark:
> Why should this code be merged if the corresponding tools are not available in Debian Bookworm (OMV7) and Trixie (base of OMV8)?
> 
> No normal user will ever be able to make use of it because the userland apps ([bcachefs-tools package](https://packages.debian.org/sid/bcachefs-tools)) are only available in Debian Sid, which OMV does not make use of.

Yes the goal is to start a work of support of bcachefs and if necessary I can make a package bcachefs-tools.deb for your repo.
Which will be compatible and accessible to all users easily.

votdev:
> > Why should this code be merged if the corresponding tools are not available in Debian Bookworm (OMV7) and Trixie (base of OMV8)?
> > No normal user will ever be able to make use of it because the userland apps ([bcachefs-tools package](https://packages.debian.org/sid/bcachefs-tools)) are only available in Debian Sid, which OMV does not make use of.
> 
> Yes the goal is to start a work of support of bcachefs and if necessary I can make a package bcachefs-tools.deb for your repo. Which will be compatible and accessible to all users easily. PS : Also I don't use debian sid and won't or need to. All work is done on debian 12

I would suggest to schedule this feature to OMV8 because adding a backported package to the stable code branch is too dangers because this code path is always executed when filesystems are processed. In OMV8 we have some time to find bugs and react on them; in addition, problems in a new version are bearable if it is still in the test cycle.

ryecoaaron:
Even on OMV 8, bcachefs will still be an experimental filesystem.  Being included directly in OMV makes me worry that noobs will use it.  Seems like it should at least be a plugin.

votdev:
> Even on OMV 8, bcachefs will still be an experimental filesystem. Being included directly in OMV makes me worry that noobs will use it. Seems like it should at least be a plugin.

That’s what I was also thinking about.

ryecoaaron:
With bcachefs being marked as externally maintained now, the dkms package should be used.  bcachefs does have an apt repo with trixie support but this would need to be added before using the bcachefs features.
https://apt.bcachefs.org/

votdev:
First of all, thank you for the code contribution. Is it possible to make this a plugin, like the bcache plugin? Since not everyone needs this feature, it makes sense not to include this functionality in the base package and instead deliver it as an upgradeable plugin.

You can use the following plugins as template:
- https://github.com/openmediavault/openmediavault/tree/master/deb/openmediavault-md
- https://github.com/openmediavault/openmediavault/tree/master/deb/openmediavault-lvm2
- https://github.com/OpenMediaVault-Plugin-Developers/openmediavault-zfs
- https://github.com/OpenMediaVault-Plugin-Developers/openmediavault-mergerfs

You can also make the [openmediavault-apt](https://github.com/openmediavault/openmediavault/tree/master/deb/openmediavault-apt) plugin a dependency of your plugin and ship it with a default configuration for the `apt.bcachefs.org` package repository.

liberodark:
> First of all, thank you for the code contribution. Is it possible to make this a plugin, like the bcache plugin? Since not everyone needs this feature, it makes sense not to include this functionality in the base package and instead deliver it as an upgradeable plugin.
> 
> You can use the following plugins as template:
> 
>     * https://github.com/openmediavault/openmediavault/tree/master/deb/openmediavault-md
> 
>     * https://github.com/openmediavault/openmediavault/tree/master/deb/openmediavault-lvm2
> 
>     * https://github.com/OpenMediaVault-Plugin-Developers/openmediavault-zfs
> 
>     * https://github.com/OpenMediaVault-Plugin-Developers/openmediavault-mergerfs
> 
> 
> You can also make the [openmediavault-apt](https://github.com/openmediavault/openmediavault/tree/master/deb/openmediavault-apt) plugin a dependency of your plugin and ship it with a default configuration for the `apt.bcachefs.org` package repository.

Hi,

I need to take time on this.
Thanks for your suggestion.

Best Regards

ryecoaaron:
Starting with the code here, I created a bcachefs plugin and added a ton of features - https://github.com/OpenMediaVault-Plugin-Developers/openmediavault-bcachefs

votdev:
Closing because the code has been refactored as a plugin.

PR #1938, opened 2025-03-27
fix: locale/zh_CN: Fix a typo "SHH"

Avenger-285714:
"""
The context indicates that "SHH" should be "SSH" based on the intended meaning.

Correct this typo.

Signed-off-by: WangYuli <wangyuli@uniontech.com>

"""

### DIFF
--- a/deb/openmediavault/usr/share/openmediavault/locale/zh_CN/openmediavault.po
+++ b/deb/openmediavault/usr/share/openmediavault/locale/zh_CN/openmediavault.po
@@ -1,7 +1,7 @@
 # SOME DESCRIPTIVE TITLE.
 # Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
 # This file is distributed under the same license as the openmediavault package.
-# 
+#
 # Translators:
 # Herald Yu <yuhr123@gmail.com>, 2018-2019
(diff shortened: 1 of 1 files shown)


votdev:
Thx for the fix. I've applied the fix at Transifex (https://translate.openmediavault.org), the translation service used by OMV. Translations are only pulled into the git repos, so translations must be done at Transifex.

Avenger-285714:
> Thx for the fix. I've applied the fix at Transifex (https://translate.openmediavault.org), the translation service used by OMV. Translations are only pulled into the git repos, so translations must be done at Transifex.

Okay, thank you. In addition, I've already requested to join the openmediavault project on Transifex. My Transifex ID is Avenger_285714.

votdev:
> > Thx for the fix. I've applied the fix at Transifex ([translate.openmediavault.org](https://translate.openmediavault.org)), the translation service used by OMV. Translations are only pulled into the git repos, so translations must be done at Transifex.
> 
> Okay, thank you. In addition, I've already requested to join the openmediavault project on Transifex. My Transifex ID is Avenger_285714.

Please check https://docs.openmediavault.org/en/stable/development/contribute.html#how-to-become-a-translator how to become a translator. Thx.

Issue #1170, opened 2021-12-28
Add reload on change modifier


ryecoaaron:
**Is your feature request related to a problem? Please describe.**
I need to create cascading dropdowns - populate a child dropdown from the value selected from a parent dropdown.

**Describe the solution you'd like**
Add a modifier to reload on change of a specified prop.
Add ability to use the value of a field in the proxy for rpc calls.

**Describe alternatives you've considered**
Fully populate all dropdowns with all possible values and throw errors on the rpc side when values of a parent and child dropdown have incompatible values.

**Additional context**
example code:

```
      - type: select
        name: arch
        label: _("Architecture")
        value: "x86_64"
        textField: arch
        valueField: arch
        store:
          proxy:
            service: 'Kvm'
            get:
              method: 'enumerateArchitectures'
      - type: select
        name: cpu
        label: _("CPU")
        value: 'host-passthrough'
        textField: cpu
        valueField: cpu
        store:
          proxy:
            service: 'Kvm'
            get:
              method: 'enumerateCpusByArch'
              params:
                arch: '{{ arch }}'
        modifiers:
          - type: reload
            constraint:
              operator: change
              arg0:
                prop: arch
```

votdev:
I'm sorry, didn't find a way to implement that

votdev:
I'm sorry, there is no way to implement that :-(

PR #1945, opened 2025-04-05
Add support to dynamically reload `select` form fields

votdev:
A reload is triggered when the `constraint` of a modifier is true or the form fields specified in `deps` have been modified. The specified request parameters (DataStore) are used to reload the content via RPC.

The following code will trigger a `reload` when the field `disallowusermod` is modifed. Instead of `deps` you can also use `constraint`.
```
{
        type: 'select',
        name: 'shell',
        label: gettext('Shell'),
        placeholder: gettext('Select a shell ...'),
        value: '/usr/bin/sh',
        store: {
          proxy: {
            service: 'System',
            get: {
              method: 'getShells'
            }
          },
          sorters: [
            {
              dir: 'asc',
              prop: 'text'
            }
          ]
        },
        modifiers: [
          {
            type: 'reload',
            typeConfig: {
              proxy: {
                service: 'System',
                get: {
                  method: 'getShells2'
                }
              },
              sorters: [
                {
                  dir: 'desc',
                  prop: 'text'
                }
              ]
            },
            deps: ['disallowusermod']
          }
        ]
      },
```

Replace the via RPC loaded options with static data.
```
{
        type: 'select',
        name: 'shell',
        label: gettext('Shell'),
        placeholder: gettext('Select a shell ...'),
        value: '/usr/bin/sh',
        store: {
          proxy: {
            service: 'System',
            get: {
              method: 'getShells'
            }
          },
          sorters: [
            {
              dir: 'asc',
              prop: 'text'
            }
          ]
        },
        modifiers: [
          {
            type: 'reload',
            typeConfig: {
              data: ['/usr/bin/foo']
            },
            deps: ['disallowusermod']
          }
        ]
      }
```

Fixes: https://github.com/openmediavault/openmediavault/issues/1170

- [x] References issue

### DIFF
--- a/deb/openmediavault/debian/changelog
+++ b/deb/openmediavault/debian/changelog
@@ -1,6 +1,7 @@
 openmediavault (7.7.4-1) stable; urgency=medium
 
-  *
+  * Issue #1170: Add support to dynamically reload `select` form
+    fields.
 
  -- Volker Theile <volker.theile@openmediavault.org>  Sat, 05 Apr 2025 12:50:50 +0200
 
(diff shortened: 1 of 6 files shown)


ryecoaaron:
Thank you for adding this.  I have wanted cascading dropdowns for a long time.

votdev:
> Thank you for adding this. I have wanted cascading dropdowns for a long time.

I don't know why I didn't come up with the solution to implement this feature earlier. :-)

PR #1946, opened 2025-04-11
Updated NL translation

DartAssan:
Updated NL translation: some fixes + non-existent ones were added


votdev:
Thank you for this contribution. I have to close it because translations are managed via [Transifex](https://transifex.com) and synced from there from time to time.
I've imported the file there. If you want to plan to do more translations, please check https://docs.openmediavault.org/en/stable/development/contribute.html#how-to-become-a-translator.

PR #1947, opened 2025-04-16
Display only relevant iface config

votdev:
Additionally add the `override` nunjucks filter to manipulate a piped value if the specified condition is truthy.

- [ ] References issue


votdev:
Will be part of `openmediavault 7.7.5`.

Issue #1673, opened 2023-12-29
NFS Share improved UI documentation / regex check


HomeLabineer:
**Please only open an issue here, when advised by a moderator in openmediavault forum.** - yea, no - see my comment on this down below

Requests for support and questions shall be addressed first to openmediavault forum: https://forum.openmediavault.org/

**Note:** Issues for non-core plugins are to be raised in the corresponding https://github.com/OpenMediaVault-Plugin-Developers repository

**Describe the bug**

NFS sharing page let's me attempt to save config w/ invalid subnet list (not white-space separated)

**To Reproduce**

1. Go to 'services > nfs > shares'
2. Add or edit an existing share
3. Save it
4. Apply it

Received the following:
```
----------
          ID: start_nfs_server_service
    Function: service.running
        Name: nfs-server
      Result: False
     Comment: Job for nfs-server.service canceled.
     Started: 08:50:44.706775
    Duration: 171.948 ms
     Changes:
```

Checked a couple things, found this:

```
root@vm-omv-01:~# sudo systemctl status nfs-server.service
● nfs-server.service - NFS server and services
     Loaded: loaded (/lib/systemd/system/nfs-server.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Fri 2023-12-29 08:50:44 CST; 3min 21s ago
    Process: 13937 ExecStartPre=/usr/sbin/exportfs -r (code=exited, status=1/FAILURE)
    Process: 13938 ExecStopPost=/usr/sbin/exportfs -au (code=exited, status=0/SUCCESS)
    Process: 13939 ExecStopPost=/usr/sbin/exportfs -f (code=exited, status=0/SUCCESS)
        CPU: 10ms

Dec 29 08:50:44 vm-omv-01 exportfs[13937]: exportfs: Invalid prefix `24,' for 192.168.11.0/24,
Dec 29 08:50:44 vm-omv-01 exportfs[13937]: exportfs: No options for /export 192.168.11.0/24,: suggest 192.168.11.0/24,(sync) to avoid warning
Dec 29 08:50:44 vm-omv-01 exportfs[13937]: exportfs: /etc/exports [3]: Neither 'subtree_check' or 'no_subtree_check' specified for export "192.168.11.0/24,:/export".
Dec 29 08:50:44 vm-omv-01 exportfs[13937]:   Assuming default behaviour ('no_subtree_check').
[...]

PR #1951, opened 2025-05-05
Add a warning to NFS client field about multiple entries

KucharczykL:
The UI for setting up NFS shares does not support multiple whitespace-separated entries as specified in the manual but at the same time the UI does not indicate this in any way.
This leads to a confusion when the NFS shares do not work at all despite the user having done nothing incorrectly.
The only way to find out the problem is analyzing journal messages (see below) which indicate that the syntax of the `/etc/exports` file is incorrect.

This has been discussed before and regex solution has been rejected due to being too complex.
I propose to add a simple message in the UI that would warn the user that multiple entries are not supported.

```
May 05 12:37:06 nas systemd[1]: Starting nfs-server.service - NFS server and services...
May 05 12:37:06 nas exportfs[2851334]: exportfs: No options for /export/nas 192.168.0.0/24: suggest 192.168.0.0/24(sync) to avoid warning
May 05 12:37:06 nas exportfs[2851334]: exportfs: /etc/exports [2]: Neither 'subtree_check' or 'no_subtree_check' specified for export "192.168.0.0/24:/export/nas".
May 05 12:37:06 nas exportfs[2851334]:   Assuming default behaviour ('no_subtree_check').
May 05 12:37:06 nas exportfs[2851334]:   NOTE: this default has changed since nfs-utils version 1.0.x
```

```
cat /etc/exports
# This file is auto-generated by openmediavault (https://www.openmediavault.org)
# WARNING: Do not edit this file, your changes will get lost.

# /etc/exports: the access control list for filesystems which may be exported
#               to NFS clients.  See exports(5).
/export/nas 192.168.0.0/24 100.64.0.0/10(fsid=f8ed4e05-52f1-46d8-aa57-4ac6a404e57a,rw,subtree_check,insecure)
```

References: 
- https://github.com/openmediavault/openmediavault/issues/1673
- https://forum.openmediavault.org/index.php?thread%2F53016-listing-multiple-nfs-clients-can-be-confusing-to-the-user%2F

### DIFF
--- a/deb/openmediavault/workbench/src/app/pages/services/nfs/nfs-share-form-page.component.ts
+++ b/deb/openmediavault/workbench/src/app/pages/services/nfs/nfs-share-form-page.component.ts
@@ -69,7 +69,9 @@ export class NfsShareFormPageComponent extends BaseFormPageComponent {
           gettext(
             // eslint-disable-next-line max-len
             "Please check the <a href='https://manpages.debian.org/nfs-kernel-server/exports.5.html' target='_blank'>manual page</a> for more details."
-          ),
+          ) +
+          ' ' +
+          gettext("WARNING: UI only supports a single client entry per share."),
         value: '',
         validators: {
           required: true


ryecoaaron:
While the field is called "Client", I think some people might read "single client" and nothing else.  This might make them think only one client can connect to the share.  I don't know how to make it more clear.

deb/openmediavault/workbench/src/app/pages/services/nfs/nfs-share-form-page.component.ts
```diff
@@ -69,7 +69,9 @@ export class NfsShareFormPageComponent extends BaseFormPageComponent {
           gettext(
             // eslint-disable-next-line max-len
             "Please check the <a href='https://manpages.debian.org/nfs-kernel-server/exports.5.html' target='_blank'>manual page</a> for more details."
-          ),
+          ) +
+          ' ' +
+          gettext("WARNING: UI only supports a single client entry per share."),
```
votdev, line 74: This is incorrect. You can limit the number of clients/hosts via prefix length.

1 host: 192.0.2.0/30
~100 hosts: 192.0.2.0/25
255 host: 192.0.2.0/24

KucharczykL: I tried to get that point across by stating "single client entry" instead of "single client" but I guess that's too easy to overlook/misunderstand. What about "UI does not support multiple comma-separated values", then?

votdev: > What about "UI does not support multiple comma-separated values", then?

This issue is not only related to the UI. This is how the configuration pipeline of OMV works for NFS.

votdev: And this is not a `WARNING`, it's only a `NOTE`.

KucharczykL: > Perhaps `Note: Only one host can be configured per share` is more appropriate.

You said yourself that 192.0.2.0/25 is ~100 hosts so this does not seem to clear up the confusion at all?

> > What about "UI does not support multiple comma-separated values", then?
> 
> This issue is not only related to the UI. This is how the configuration pipeline of OMV works for NFS.

Let's focus on OMV, then, by changing "UI" to "OMV". From the point of view of the user, the NFS manual (linked from the field text) says whitespace-separated values are supported when in fact they are not supported by OMV. This is the source of confusion.

votdev: `host` is the name used in the man page for describing how the /etc/exports line is constructed.

> Let's focus on OMV, then, by changing "UI" to "OMV".

This is totally superfluous because the user is in the OMV UI. Therefore it is not necessary to mention this again.

KucharczykL: It's true that the user is in the OMV UI but at the same time the user is also reading a hint text that mentions the NFS manual. It needs to be clear that there is discrepancy between what NFS supports and what OMV supports (via the Client field) so saying "OMV UI" in this case is not superfluous.

votdev:
Thx for working on this.

I am closing this PR because it is superseded by https://github.com/openmediavault/openmediavault/commit/a20af669533a60c2b8f392872ac1b10194619395.

KucharczykL:
Maybe I am bad at explaining myself but this _does not_ clear up the confusion, it only adds to it because you are using terms as defined in the man page but the user does not know that so they will just understand them as they are generally used.

votdev:
> Maybe I am bad at explaining myself but this _does not_ clear up the confusion, it only adds to it because you are using terms as defined in the man page but the user does not know that so they will just understand them as they are generally used.

But then the problem lies with the user. If I introduce my own terms here, it won't help. We saw that in the discussion above. The terms used are those used in the manual. In my opinion, anything else leads to even more confusion.

PR #1952, opened 2025-05-05
Fix parent device selection for NVMe devices

xOneca:
`nvme0n1p1` (partition) and `nvme0n1` (namespace) are block devices but `nvme0` (controller) is not a block device, even if it also has a device node.

This fixes the case when a filesystem resides in an unpartitioned namespace and the parent controller is returned instead of the namespace device itself.

### DIFF
--- a/deb/openmediavault/usr/lib/python3/dist-packages/openmediavault/fs/__init__.py
+++ b/deb/openmediavault/usr/lib/python3/dist-packages/openmediavault/fs/__init__.py
@@ -220,6 +220,7 @@ class Filesystem(openmediavault.device.BlockDevice):
         * /dev/cciss/c0d0p2 => /dev/cciss/c0d0
         * /dev/md0 => /dev/md0
         * /dev/dm-0 => /dev/dm-0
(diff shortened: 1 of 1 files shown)


votdev:
How to reproduce this issue? Where is the Python code involved?

Does the PHP equivalent code handle this case correctly?

votdev (approved):
LGTM

# Unpatched
```
Python 3.11.2 (main, Nov 30 2024, 21:22:50) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import openmediavault.fs
>>> fs = openmediavault.fs.Filesystem("/dev/nvme0n1")
>>> fs.parent_device_file
'/dev/nvme0'
```

# Patched
```
Python 3.11.2 (main, Nov 30 2024, 21:22:50) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import openmediavault.fs
>>> fs = openmediavault.fs.Filesystem("/dev/nvme0n1")
>>> fs.parent_device_file
'/dev/nvme0n1'
```

votdev:
@xOneca Thx

PR #1970, opened 2025-06-05
FR: Add UDEV rule to fix the duplicate serial number issue for USB enclosure

ilpianista:
Add UDEV rule to fix the duplicate serial number issue for External SATA Hard Drive Adapter cable PA023U3.

- [ ] References issue


votdev:
Thx

Issue #1972, opened 2025-06-17
k8s certificates assigned through settings are overwritten by cert-manager


keif888:
**Describe the bug**

When assigning a certificate to the k8s plugin via the setting page, the certificate is applied by the omv-salt command, but, the cert-manager then detects that the default-tls-cert secret has changed, and regenerates it.
This prevents the OMV assigned certificate from being used with the kubernetes hosted apps.

**To Reproduce**

1. Generate certificates to use with Open Media Vault (I used OpenSSL)
2. Add the certificates to Open Media Vault
2.1. Open System, Certificates, SSL
2.2. Add the certificate with the + button, uploading the PEM and KEY files
3. Install the Kubernetes plugin
4. Open Services, Kubernetes, Settings
5. Configure the Certificate option, with the certificates added in 2.2 above
6. Save the settings, and deploy
7. Wait for Kubernetes to be installed and configured
8. Click on the Open UI button
9. See certificate that is self generated by Kubernetes

**Expected behavior**
The certificate that is used for the Kubernetes UI should be the certificate that was configured in step 2.2 above.

**Screenshots**

![Image](https://github.com/user-attachments/assets/6c4b1237-4021-4d6f-9a7f-3650d0331e68)

**Reference to Forum**

[URL link to forum post](https://forum.openmediavault.org/index.php?thread/56486-kubernetes-custom-certificate-management-e-g-letsencrypt/)

**openmediavault Server (please complete the following information):**

- OS version: 
Linux nas 6.12.27+bpo-amd64 # 1 SMP PREEMPT_DYNAMIC Debian 6.12.27-1~bpo12+1 (2025-05-19) x86_64 GNU/Linux
- openmediavault version: 
ii  openmediavault                   7.7.9-1                              all          openmediavault - The open network attached storage solution
ii  openmediavault-k8s               7.4.9-1                              all          openmediavault Kubernetes plugin
ii  openmediavault-keyring           1.0.2-2                              all          GnuPG archive keys of the openmediavault archive
ii  openmediavault-omvextrasorg      7.0.2                                all          OMV-Extras.org Package Repositories for OpenMediaVault
ii  openmediavault-zfs               7.1.4                                amd64        OpenMediaVault plugin for ZFS

**Client (please complete the following information):**

- Device: amd64 Computer
- OS: Windows 11, Debian, Ubunto
- Browser: Chrome/Edge
- Version: 137.0.7151.104/137.0.3296.83 

**Additional context**

[...]

PR #1973, opened 2025-06-17
fix: k8s certificates assigned through settings are overwritten by cert-manager

keif888:
Fixes: #1972

- [X] References issue
- [X] Includes tests for new functionality or reproducer for bug

**To Reproduce**

1. Generate certificates to use with Open Media Vault (I used OpenSSL)
2. Add the certificates to Open Media Vault
2.1. Open System, Certificates, SSL
2.2. Add the certificate with the + button, uploading the PEM and KEY files
3. Install the Kubernetes plugin
4. Open Services, Kubernetes, Settings
5. Configure the Certificate option, with the certificates added in 2.2 above
6. Save the settings, and deploy
7. Wait for Kubernetes to be installed and configured
8. Click on the Open UI button
9. See certificate that is self generated by Kubernetes

This change uses a kubernetes secret openmediavault-tls-cert when a certificate is provided in the k8s settings certificate option for the traefik TLSStore.  Otherwise it uses the default-tls-cert (current behaviour).
The secret is created within the existing file /var/lib/rancher/k3s/server/manifests/openmediavault-misc.yaml instead of the default-tls-cert.

### DIFF
--- a/deb/openmediavault-k8s/debian/changelog
+++ b/deb/openmediavault-k8s/debian/changelog
@@ -4,6 +4,7 @@ openmediavault-k8s (7.4.10-1) stable; urgency=medium
   * Fix the file mode bits of the systemd unit file k3s.service to prevent
     "is marked world-writable" warnings.
   * Disable Kong admin UI which is part of the Kubernetes Dashboard.
+  * Issue #1972: Fix bug when using user defined SSL certificates.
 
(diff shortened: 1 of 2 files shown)


votdev:
Will review it asap.

votdev:
@keif888 What are you thinking about my changes?

keif888:
Looks good.
I didn't realise there was an else in the scripting language.

votdev:
Thx

PR #1974, opened 2025-06-23
Minor readme improvements for vagrant windows setup

tas50:
Typo + some markdown fixes for the Vagrant Windows readme

### DIFF
--- a/vagrant/windows/README.md
+++ b/vagrant/windows/README.md
@@ -1,13 +1,16 @@
 # Windows test environment
 
-This Vagrant box can be used to test various features like SMB `Previos Versions` which are not available on Linux.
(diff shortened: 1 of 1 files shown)


vagrant/windows/README.md
```diff
@@ -2,15 +2,13 @@
...
-## Requirements
-
 1. Install plugins
 
 ```
-vagrant plugin install winrm
+$ vagrant plugin install winrm
 ```
 
-## Troubleshooting
```
tas50: Do you mind if I fix this in both? There should only be 1 H1 in a markdown doc

votdev: In my opinion, this makes no sense here as there is no real main title in these documents.

PR #1976, opened 2025-06-23
Update all GNU license headers to use https links

tas50:
Use the HTTPS link to gnu.org in all license headers


votdev:
I've applied the changes in a separate commit because reviewing this PR was too time consuming using the GH UI.

I hope you understand that I can't merge a PR without a review; I just want to avoid overlooking code in the masses that doesn't belong there.

votdev:
Anyway, thx for your contribution.

PR #1985, opened 2025-07-12
Wrap fields in a form container

votdev:
- Remove `flex` property from form fields.
- The width of the individual fields in a container is automatically adjusted to the number of form fields it contains. All fields have the same width.
- Fields are wrapped to the next line if the width of the browser window is not wide enough to display all fields in one line. The display on small screens such as tablets benefits from this.

Plugins do not have to adapt their code; the `flex` form field property is simply ignored.

### Wide viewport
grafik

### Small viewport
grafik

### Other
grafik

grafik


ryecoaaron:
Sorry, I wasn't about to review as I was on vacation.  Seems fine to me.

Issue #2001, opened 2025-08-16
Network stats shifted


ranger81:
Hi,

I noticed an error in the network statistics.

The regex in line [424](https://github.com/openmediavault/openmediavault/blob/35038aee8aba2b7d9deb029292e616ae7c9674cb/deb/openmediavault/usr/share/php/openmediavault/system/net/networkinterface.inc#L424C28-L424C29)

`$regex = sprintf('/^\s*%s:(.*)$/im', $this->getDeviceName());`

should be replaced with

`$regex = sprintf('/^\s*%s:[\s]?(.*)$/im', $this->getDeviceName());`

to also match correctly a line where there is a space between the interface name and the rx bytes. Otherwise all stats outputs are shifted.

PR #2002, opened 2025-08-16
Fixed network statistics #2001

ranger81:
Fix for the shifted network statistics

In case the /proc/net/dev output line has a space between the iface name and the rx bytes, the existing regex does not match properly. This pull request should fix the issue #2001 .

Fixes: #2001 

Signed-off-by: Tobias Link <ranger81@dsuclan.de>


ranger81:
Hi,
Sorry this is my first commit to this repository and I messed up the Singed-Off-by stuff, my apologies!
Hope you can still review the PR.
Thanks

votdev:
Wouldn’t it be better to trim the content in matches. The regex gets more complex with your approach to simply get rid of blanks?

$data = preg_split("/[\s]+/", $matches[1]);

to

$data = preg_split("/[\s]+/", trim($matches[1]));

in line [429](https://github.com/openmediavault/openmediavault/blob/35038aee8aba2b7d9deb029292e616ae7c9674cb/deb/openmediavault/usr/share/php/openmediavault/system/net/networkinterface.inc#L429).

ranger81:
I think this would be a good alternative solution as well, yes.

votdev:
Thanks for the fix. I've implemented my comment via 6a0cb9250db5bc61f1f0d57987e82d0f7bf68fd0 because this is IMO the better approach and this PR is missing the signed-off-by line in the commit message.

PR #2034, opened 2025-10-26
Fixing cases of wrongly assumed temperature values. 

robNice:
This fixes a failure using smartctl to get hdd temperature values relied on the related smart IDs.

Usually the IDs 190, 194 and 231 where used for temperature values. Some SandForce or Phison controlled SSDs, like the Corsair Force LS SSD, don't have any temperature sensors at all. This can lead to finding a wrong attribute at ID 231, like "SSD_LIFE_LEFT", giving the maintainer slight heart attacks.

getTemperature() used to solely check for one of the three IDs. As, as far as my researches were correct, the attribute_name is always a string containing the substring "Temperature". Including a check for this substring should do the trick. 

Fixes: 
There are two blocks inside the getTemperature() method to find the right temperature value. One for smart devices and one for non-smart devices. This fixes the smart-device-block.
 - ditching the switch case for a short in_array()
 - adding a second comparison to check if the attribute name does include the substring "temperature" (case insensitive)
 - moving the comment blocks inside the switch cases to an area right before the foreach-loop.

Signed-off-by: robNice <robnice@downlord.net>

### DIFF
--- a/deb/openmediavault/usr/share/php/openmediavault/system/storage/smartinformation.inc
+++ b/deb/openmediavault/usr/share/php/openmediavault/system/storage/smartinformation.inc
@@ -451,41 +451,46 @@ class SmartInformation {
 			return FALSE;
 		$result = 0;
 		$found = FALSE;
-		// Process the attributes to get the temperature value.
+
+		// Examples of the smartctl result structure of the possible
+		// IDs (#190, #194, #231) that are representing temperature values
+		//
+		// ID# ATTRIBUTE_NAME          FLAG     VALUE WORST THRESH TYPE      UPDATED  WHEN_FAILED RAW_VALUE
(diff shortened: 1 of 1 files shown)


votdev:
Already fixed with f50bbed844608df9e5a358f4d2273a0d69de7965, 33f7113059c2a90fdbc82ce667952ca11bd58165, f1b053d0ae876db726d62fa1bb0a03ab38412ed9 and eb1f1be0f4616ba89c9e29246c4ea6adaa3b6ae5. Nevertheless, thank you for your contribution.

robNice:
> Already fixed with [f50bbed](https://github.com/openmediavault/openmediavault/commit/f50bbed844608df9e5a358f4d2273a0d69de7965) and [33f7113](https://github.com/openmediavault/openmediavault/commit/33f7113059c2a90fdbc82ce667952ca11bd58165). Nevertheless, thank you for your contribution.

lol - I just noticed you were 2 minutes faster than me.

PR #2046, opened 2025-11-18
Fix typo.

reitermarkus:
Just noticed this while adding some cron jobs.


votdev:
BTW, you do not need to modify the *.po files in a PR. They are never pushed to the translation service at https://translate.openmediavault.org. The translation files are only pulled from there. The translations need to be modified there. The only file that needs to be re-build is the *.pot file which is pushed to the translation service then.

PR #2053, opened 2025-11-24
Add functionality to execute a function to populate an environment variable

votdev:
Note, the `OMV_ROOT_DEVICE_FILE` is not really necessary because `\OMV\System\System::getRootDeviceFile()` is already caching the result, but it is a nice example how the new feature works.

- [ ] References issue


votdev:
@ryecoaaron This feature can be useful, for example, to determine whether an application is installed or not when omv-engined is started. Then this does not have to be checked constantly during runtime, e.g. when a RPC is executed. Of course, it is only really useful if it concerns things that are not changed at any point after omv-engined is started.

PR #2065, opened 2025-11-30
Add IPv6 detection when installing OMV over Debian

UltimateByte:
Added IPv6 support to [40interfaces.sh](https://github.com/openmediavault/openmediavault/blob/master/deb/openmediavault/usr/share/openmediavault/confdb/populate.d/40interfaces.sh).

Initial script skipped IPv6 detection.
It now detects IPv6.

Fixes: https://forum.openmediavault.org/index.php?thread/57706-providing-help-and-getting-eta-for-omv8/&postID=427168#post427168

First PR, signed and sent the OCA.
I hope everything is in place, please tell me if not.

Signed-off-by: Robin LABADIE - LRob <git@lrob.fr>

### DIFF
--- a/deb/openmediavault/usr/share/openmediavault/confdb/populate.d/40interfaces.sh
+++ b/deb/openmediavault/usr/share/openmediavault/confdb/populate.d/40interfaces.sh
@@ -38,18 +38,21 @@ for dnsnameserver in $(grep -iP "^\s*dns-nameservers\s+.*$" ${OMV_INTERFACES_CON
 done
 
 # Configure ethernet network interfaces.
-# Understanding systemd’s predictable network device names:
+# Understanding systemd's predictable network device names:
 # https://github.com/systemd/systemd/blob/master/src/udev/udev-builtin-net_id.c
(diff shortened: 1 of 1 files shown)


votdev:
What happens if there is no IPv4 configuration for an interface and only IPv6? I think the code will not cover that.

votdev:
Please add the `Signed-off-by: Robin LABADIE <git@lrob.fr>` line to the commit and not the PR description, otherwise the DCO check will block this PR. Have a look into https://github.com/openmediavault/openmediavault/blob/master/CONTRIBUTING.md.

UltimateByte:
Good catch, if we handle IPv6, we'd better handle IPv6-only too! I'll add this.

I'll also put my info inside the commit (of course there had to be at least one mistake for my 1st PR! haha)

Though "Make sure your PR has only one commit."
I would assume I need to redo my branch with only that one commit, unless you can "squash and merge" from your end.
Your call, should I make another branch then redo a new PR?
(meanwhile I'll add the commit to my current branch for feedback anyways)

votdev:
Just out of interest, why don't you extract the static IPv6 configuration from `/etc/network/interfaces` and use the `ip` command instead?

PR #2083, opened 2025-12-27
Add email notification batching for scheduled power operations

### LINKED ISSUE #2084: Too many emails during scheduled reboots/shutdowns
## Too many emails during scheduled reboots/shutdowns

  Right now when you schedule a reboot or shutdown, you get spammed with individual emails for every single service that stops or starts. A simple reboot can easily trigger 10-20 separate emails.

  This is annoying and makes it hard to see what actually happened.

  ## What I'd like instead

  Batch all those notifications into one email. Something like:

  Subject: Scheduled REBOOT - 12 service events

  Your scheduled reboot triggered these events:

  [10:30:15] nginx: stopped
  [10:30:18] php-fpm: stopped
  [10:30:45] collectd: stopped
  [10:31:20] collectd: started
  [10:31:22] nginx: started
  ...

  ## How it could work

  - Add an option in System → Power Management → Settings to enable batching
  - For shutdown/standby: wait 60 seconds to collect events, send one email, then shut down
  - For reboot: collect events during shutdown + startup, send email when system is back up
  - Only affects scheduled tasks, not manual reboots
  - Off by default so it doesn't break existing setups

  This would use the existing Monit monitoring - just batch the alerts instead of sending them individually.

Nirus2000:
Add email notification batching for scheduled power operations

  Consolidates multiple service event notifications during scheduled
  reboot/shutdown/standby into a single email. Configurable via
  System > Power Management > Settings. Disabled by default.

  - Wrapper script intercepts scheduled tasks
  - Monit alert program logs events during batching window
  - Post-reboot systemd service for reboot scenarios
  - Web UI configuration with enable toggle and time window
  
  Closes #2084

### DIFF
--- a/deb/openmediavault/debian/openmediavault.postinst
+++ b/deb/openmediavault/debian/openmediavault.postinst
@@ -201,6 +201,7 @@ case "$1" in
 		deb-systemd-helper enable openmediavault-cleanup-php.service
 		deb-systemd-helper enable openmediavault-engined.service
 		deb-systemd-helper enable openmediavault-issue.service
+		deb-systemd-helper enable openmediavault-powermgmt-postreboot.service
 
 		########################################################################
 		# Activate trigger to rebuild workbench configuration files.
--- a/deb/openmediavault/etc/systemd/system/openmediavault-powermgmt-postreboot.service
+++ b/deb/openmediavault/etc/systemd/system/openmediavault-powermgmt-postreboot.service
@@ -1,0 +1,21 @@
+[Unit]
+Description=Send consolidated power management notifications after reboot
+After=network-online.target monit.service openmediavault-engined.service
+Wants=network-online.target
+
+[Service]
+Type=oneshot
+ExecStartPre=/bin/bash -c 'if [ -f /var/lib/openmediavault/powermgmt-operation.state ]; then \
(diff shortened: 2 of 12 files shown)


votdev:
TBH, i don't like the complexity of this feature and the PR. That seems a little too complicated to me for a problem that probably only affects a tiny fraction of users. The implementation smacks of maintenance hell. Since I have seen your other GH issue, I think I know which direction this is going. As already described in the rejected issue, the idea does not fit with the project's objective, which is aimed at home users and small office installations.

votdev:
Could you please open an issue that describes the problem and the solution.

Nirus2000:
Ref.: #2084

votdev:
I'm sorry, but I have to close this PR. This feature does not fit with the project goals, which include keeping things as simple as possible. I'm sorry that you've already invested time in the PR. But https://github.com/openmediavault/openmediavault/blob/master/CONTRIBUTING.md explicitly states that an issue should be opened first to discuss things.

votdev:
A first step would be to check if the https://github.com/openmediavault/openmediavault/blob/master/deb/openmediavault/debian/openmediavault.openmediavault-cleanup-monit.service is still necessary and whether removing this cleanup service will partly fix the behaviour.

PR #2096, opened 2026-01-10
Use more secure permissions for the sshd config file

tas50:
Using the CIS benchmark recommendation that sshd should be 600. No need for anyone but root to be reading this.

Signed-off-by: Tim Smith <tsmith84@gmail.com>

### DIFF
--- a/deb/openmediavault/srv/salt/omv/deploy/ssh/20sshd_config.sls
+++ b/deb/openmediavault/srv/salt/omv/deploy/ssh/20sshd_config.sls
@@ -33,7 +33,7 @@ configure_sshd_config:
         config: {{ config | json }}
     - user: root
     - group: root
-    - mode: 644
+    - mode: 600
 
 divert_sshd_config:
   omv_dpkg.divert_add:


votdev:
The Debian default values are 644. the file does not contain any sensitive data which would justify to use more conservative permissions.
Do you have any references (including URLs) that have strong reasons to use 600?

tas50:
Here's the policy from the latest CIS Debian 13 benchmark:

image
image

votdev:
> Configuration specifications for sshd need to be protected from unauthorized changes
by non-privileged users.

644 is fulfilling that. The reference does not mention WHY non-privileged users should not read it.

I know this is pedantic, but when I make changes, I want to see reasons why it makes sense. Doing it because someone with a grand and important name (here CIS) says so is not enough.

tas50:
I'm not sure how the download works without membership, but here's the link they claim lets you download the complete benchmarks: https://www.cisecurity.org/benchmark/debian_linux

votdev:
This is what Claude thinks about it:

> some administrators prefer more restrictive permissions like 600 (rw—––) to prevent non-root users from reading the configuration, though this isn’t strictly necessary from a security standpoint since the file doesn’t contain private keys or passwords.​​​​​​​​​​​​​​​​

tas50:
That's fair. My end goal was to get my system CIS Level 1 compliant. They 100% have not provided a lot of actual justification for this control.

PR #2099, opened 2026-01-12
Be more consistent with default ACL of shared folders

lisanet:
Be more consistent with default acl: 

1- create shared folder -> no default ACL
2- change file permson ACL-Tab -> no default ACL added 
3- define named ACL in on ACL-Tab -> default ACL + named ACL added 
4- remove/uncheck named ACL on ACL-Tab -> no default ACL + no named ACL, ACLs deleted

There's a short discussion in the forum about this patch: https://forum.openmediavault.org/index.php?thread/58142-shared-folder-standard-unix-permissions-acl-correct-behavior-in-usage-result/&postID=430310#post430310


votdev (changes requested):
Unfortunately, the code change suggested here does not work. It is not possible to unset ACL settings.

Issue #2103, opened 2026-01-15
[BUG] - overlapping text in dashboard view


Mlocik97:
**Please only open an issue here, when advised by a moderator in openmediavault forum.**

Don't have acc for OMV forum. Sorry.

**Describe the bug**

On viewport with width 1475 px, if left panel is shown, and 3 colums in dashboard are rendered, the "system time" and "uptime" are overlapping.

**To Reproduce**

1. open dashboard and add enough widgets to make it 3 colums
2. resize window to 1475 px
3. text is overlapping

**Expected behavior**

system time could be rendered in two lines, like:

> Št 15. január 2026,
> 23:20:00

**Screenshots**

<img width="369" height="213" alt="Image" src="https://github.com/user-attachments/assets/993192cc-1757-468c-bd42-443f3d383fb3" />

**Reference to Forum**

N/A

**openmediavault Server (please complete the following information):**

- OS version: `Linux openmediavault 6.12.63+deb13-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.63-1 (2025-12-30) x86_64 GNU/Linux`
- openmediavault version:

```
ii  openmediavault                  8.0.6-2                              all          openmediavault - The open network attached storage solution
ii  openmediavault-apt              8.0-4                                all          openmediavault APT management plugin
ii  openmediavault-clamav           8.0-4                                all          openmediavault ClamAV plugin
ii  openmediavault-diskstats        8.0-4                                all          openmediavault disk monitoring plugin
ii  openmediavault-hosts            8.0-3                                all          openmediavault hosts plugin
ii  openmediavault-k8s              8.0-9                                all          openmediavault Kubernetes plugin
ii  openmediavault-keyring          1.0.2-2                              all          GnuPG archive keys of the openmediavault archive
ii  openmediavault-lvm2             8.0-3                                all          openmediavault Logical Volume Manager (LVM2) plugin
ii  openmediavault-podman           8.0-5                                all          openmediavault Podman plugin
ii  openmediavault-salt             8.0                                  amd64        Extra Python packages required by Salt on openmediavault
```

**Client (please complete the following information):**

- Device: laptop with fullHD 16:10 screen
- OS: Ubuntu 24.04.3 LTS
- Browser Chrome
- Version  143.0.7499.169 

**Additional context**

I can make the fix, and would like to work on this issue during weekend.

votdev:
Can not reproduce that issue.

# Firefox 147.0

[...]

PR #2113, opened 2026-01-21
fix overlapping text in some cases with `word-wrap`

Mlocik97:
# Description:

- fixes the bug when text in dashboard widget is overlapping.
- add unit tests

# Fixes

[#2103 - [BUG] - overlapping text in dashboard view](https://github.com/openmediavault/openmediavault/issues/2103)

- [x] References issue

### DIFF
--- a/deb/openmediavault/workbench/src/app/core/components/dashboard/dashboard-widget-system-information/dashboard-widget-system-information.component.scss
+++ b/deb/openmediavault/workbench/src/app/core/components/dashboard/dashboard-widget-system-information/dashboard-widget-system-information.component.scss
@@ -5,5 +5,14 @@
 
   .omv-grid-cell {
     break-inside: avoid;
+    overflow: hidden;
+    word-wrap: break-word;
+    word-break: break-word;
+
+    div {
+      overflow-wrap: break-word;
+      word-break: break-word;
+      word-wrap: break-word;
+    }
   }
 }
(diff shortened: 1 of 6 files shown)


votdev:
I don't know if this is better

grafik

than the original

grafik

especially when there are tooltips that display the content properly.

deb/openmediavault/workbench/src/app/core/components/dashboard/dashboard-widget-system-information/dashboard-widget-system-information.component.scss
```diff
@@ -5,5 +5,14 @@
...
   .omv-grid-cell {
     break-inside: avoid;
+    overflow: hidden;
+    word-wrap: break-word;
+    word-break: break-word;
+
+    div {
+      overflow-wrap: break-word;
+      word-break: break-word;
+      word-wrap: break-word;
```
votdev: obsolete alias for `overflow-wrap`.

deb/openmediavault/workbench/src/app/core/components/dashboard/dashboard-widget-system-information/dashboard-widget-system-information.component.spec.ts
```diff
@@ -30,4 +30,56 @@ describe('DashboardWidgetSystemInformationComponent', () => {
   it('should create', () => {
     expect(component).toBeTruthy();
   });
+
+  describe('CSS text wrapping and overflow handling', () => {
```
votdev: Nice idea, but why. These unit tests are testing things that the browser should do without worries.

Mlocik97:
Oh, I didn't check other widgets, yeah, I guess this is not good solution. Will look into it more. I had issue with testing it, so I just mocked some data in the system information widget to check if the issue was resolved.

btw. about reproducing my issue, In case you set the browser to Slovak language and system to Czech, it will be reproducible, but I checked that it renders correctly in English.

Issue #1520, opened 2023-03-27
UI: Improve user interface to support screen readers for blind users


votdev:
References:
- https://forum.openmediavault.org/index.php?thread/46139-calling-omv-developers-for-accessibility-improvement
- https://www.barrierefreies-webdesign.de/

dr-leo:
I ran into the same issue. The yellow banner for applying updates is inaccessible. My best guess is that this can be fixed very quickly by adding some alt tag to the elements generating the "buttons". Clear text instead of the graphical symbols would help as well as Jaws contains an OCR feature. 

Another workaround would be to apply the updates via a script. So is there a command to run from ssh? 

My initial overall experience is that the GUI in v6.x is more acessible to visually impaired users than in v5.x.

votdev:
I've added an aria-label to the buttons in the apply toolbar (https://github.com/openmediavault/openmediavault/commit/b65644f0c395f964a38bb1a8690a7ab84eaee66d). Hopefully this works.

Nevertheless, a full accessibility support of the UI must be provided by the community. I do not have the time to do that, the project needs my attention on too much other things.

PR #2120, opened 2026-01-26
Sync HTML lang attribute with current locale

agilegecko:
Implement success criterion [3.1.1 Language of Page (Level A)](https://www.w3.org/TR/WCAG22/#language-of-page) from [WCAG 2.2](https://www.w3.org/TR/WCAG22/) by setting HTML lang attribute with current locale.

Refers to: #1520

lang

### DIFF
--- a/deb/openmediavault/workbench/src/app/app.component.ts
+++ b/deb/openmediavault/workbench/src/app/app.component.ts
@@ -45,12 +45,15 @@ export class AppComponent implements OnInit {
   }
 
   ngOnInit(): void {
-    this.setHtmlLangAttributeFromLocale();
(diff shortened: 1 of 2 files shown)


votdev:
This will not work correct. It will not reflect the current state after a user logged in which has set a different locale. The user preferences are synced after login.

agilegecko:
> This will not work correct. It will not reflect the current state after a user logged in which has set a different locale. The user preferences are synced after login.

Here's a video of how it behaves on my machine:


https://github.com/user-attachments/assets/8ab396b7-2003-40eb-a9ae-544d27f0dfd8



Can you please explain how to reproduce the issue you're mentioning?

PR #2127, opened 2026-02-08
Add feature to configure CPU governor via UI

votdev:
Related to:
- https://forum.openmediavault.org/index.php?thread/58398-i-do-not-want-to-run-the-performance-governor-how-do-i-change-it-back
- https://forum.openmediavault.org/index.php?thread/57618-omv8-error-on-cpupower-service
- https://forum.openmediavault.org/index.php?thread/58410-error-after-update-to-8-0-10-1-cpupower-module/

- [ ] References issue


votdev:
@ryecoaaron I think these changes might affect your install script.

ryecoaaron:
install script has been updated.

PR #2128, opened 2026-02-09
Populate `conf.system.powermngmnt` at first installation

votdev:
- [ ] References issue


votdev:
@ryecoaaron Don't know if this affects you as well.

PR #2136, opened 2026-02-24
Limit systemd-journald size

votdev:
- [ ] References issue


votdev:
Is 1GiB max size of the journal enough or too small? On my productive NAS the size is ~480MiB and i see messages from 11/2025. So a retention period for about 3 months should be achievable with 1GiB.

ryecoaaron:
1GB seems fine to me.  People can set the environment variable if they want more retention.

Issue #2155, opened 2026-03-28
K8s: cannot expose same port on different protocols


ZashIn:
Adding/exposing the same port for different protocols in the Load Balancer, like 53 in TCP and UDP for DNS, results in only one being actually exposed, even though the OMV Kubernetes UI shows both being successfully added.

This seems to be [a known Kubernetes issue](https://github.com/kubernetes/kubernetes/issues/105610), which will probably not be fixed anytime soon (fix could break existing clients).
Further reference: [How a Kubernetes bug won't let you expose a service over TCP and UDP on a same port](https://ben-lab.github.io/kubernetes-UDP-TCP-bug-same-port/)

**To Reproduce**
1. Add plugin `openmediavault-k8s` with necessary config.
2. (Add a Kubernetes receipt, like a DNS server plugin Adguard Home, which requires Port 53 to be opened in TCP and UDP and configure the local DNS.)
3. Add two ports to the Load Balancer: "dns-tcp": 53 TCP and "dns-udp": 53 UDP.
4. Save and apply changes: UI shows both ports are added.
5. Check traefik service config: only one port is open (either TCP or UDP), not bot!
6. (Check DNS resolution with `dig TARGET @SERVER +tcp | +udp`: at least one request fails!)

**Expected behavior**

The Kubernetes Load Balancer UI should either:
- handle this case correctly (see workaround in the linked issue)
- or at least check if the ports are actually added correctly, ideally with an error referencing the issue above.

**Screenshots**

<img width="850" height="443" alt="Image" src="https://github.com/user-attachments/assets/d2978482-4094-4821-9356-3b2e7c0ace0b" />

traefik service yaml:
```yaml
ports:
  - name: dns-tcp
    nodePort: 31423
    port: 53
    protocol: TCP
    targetPort: dns-tcp
# Missing!
#  - name: dns-udp
#    port: 53
#    protocol: UDP
#    targetPort: dns-udp
  - name: web
    nodePort: 31511
    port: 8080
    protocol: TCP
    targetPort: web
  - name: websecure
    nodePort: 30938
    port: 8443
    protocol: TCP
    targetPort: websecure
```

**openmediavault Server (please complete the following information):**

- OS version: [`uname -a`]: `6.12.73+deb13-amd64`
- openmediavault version: [`dpkg -l | grep openmediavault`]
```
openmediavault           8.1.4-1   all     openmediavault - The open network attached storage solution
[...]

PR #2157, opened 2026-03-29
Can't expose same port on different protocols

votdev:
Fixes: https://github.com/openmediavault/openmediavault/issues/2155

References:
- https://github.com/traefik/traefik-helm-chart/pull/475
- https://github.com/traefik/traefik-helm-chart/blob/c321f599964382e882fcac19227d0483b162d2e5/traefik/values.yaml#L1076

- [x] References issue


ZashIn:
I tested `service: single: false` with adding two ports in Load Balancer settings. Should work, thank you! :+1:

Issue #2158, opened 2026-03-30
Configure apt update check frequency


mstroppel:
**Is your feature request related to a problem? Please describe.**
I'm maintaining 3 OMV servers and get almost every day apt update E-Mails. During long running operations (such as backups, transcoding...) I cannot do those updates. Also, I would be fine with a somehow "outdated" system for several days.

**Describe the solution you'd like**
I'd like to configure to do the apt update checks in a weekly or monthly manner.

**Describe alternatives you've considered**
Ignoring/deleting the E-Mails by rules could be an option.

**Additional context**
- none -

votdev:
One more thing to note: This feature does not configure how often the system checks for updates, which ultimately triggers an `apt-get update`, but rather specifies when `apticron` is called to generate a summary of the updates and send it via email. You can completely disable this email notification under `System | Notifications | Events`.

mstroppel:
Many thanks!

PR #2159, opened 2026-03-30
support configuration of apt update check frequency

mstroppel:
support configuration of apt update check frequency

Allow to configure if the apt update checks are done

- daily
- weekly
- monthly

Fixes: #2158 

Signed-off-by: Marco Stroppel <marco.stroppel@gmail.com>


mstroppel:
Hi, I did my best to get all done. Failed a bit with signed commits 🙈
I suggest the following:

1. you do not like what --> close PR and issue
2. you do not like how
    a. --> tell me please how to improve, what I need to do
    b. --> close PR (and maybe issue)

votdev:
I did some refactoring and added some missing parts that are required to get the database migration working.
I limited the schedule to `daily` and `weekly` because `monthly` is a little bit too long IMO.

I actually wanted to handle updates the same way Ubuntu does—offering them `daily`, `every two day`, `weekly` and `every two weeks`, but in the end, I was too lazy to adjust the cron configuration accordingly. This solution is better than none at all.

PR #2162, opened 2026-04-01
Set link local addresses for both Ethernet and WiFi network interfaces

votdev:
... by default to ensure that the system is reachable even if no other IP address can be obtained.

- [ ] References issue

### DIFF
--- a/deb/openmediavault/debian/changelog
+++ b/deb/openmediavault/debian/changelog
@@ -1,5 +1,8 @@
 openmediavault (8.2.1-1) stable; urgency=medium
 
+  * Set link local addresses for both Ethernet and WiFi network interfaces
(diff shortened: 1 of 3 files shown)


votdev:
The IP address (in the 169.254.0.0/16 range) is only assigned by Netplan/systemd if, for example, something goes wrong with DHCP. We can also set the link-local IPv4 address only in the case of DHCP, so that the system remains accessible even if a problem occurs. With a static configuration, this makes little sense.

votdev:
> I'm not sure why this needs to be the default but I don't have a reason to say it shouldn't be default.

You know what my philosophy, and therefore OMV’s, is: to offer a configuration that’s user-friendly and useful. Too many configuration options don’t fit with that philosophy.

PR #2194, opened 2026-05-04
Fixed filebrowser not starting in OMV8

jamfrade:
Issue described in the following forum post:
https://forum.openmediavault.org/index.php?thread/58753-omv8-file-browser-stopped-working/

Actions that fixed the issue:
Changed ports to prevent incompatibility
Added correct uid and gid so that the pod has the right privileges on monted volume

Signed-off-by: Jorge Frade <jamfrade@gmail.com>


votdev:
Thx

PR #2195, opened 2026-05-06
Remove auth from config (fix admin pass reset)

jamfrade:
Removed admin username and password configuration from config to prevent admin password reset on startup.

According documentation bellow if the admin password is set in the config file it is reset to the config value upon restart (even if changed by the user). https://filebrowserquantum.com/en/docs/configuration/authentication/password/#set-admin-password

Removed the auth part completely from the config file as it was redundant. Default user and pass are already admin:admin.

Signed-off-by: Jorge Frade jamfrade@gmail.com

### DIFF
--- a/deb/openmediavault-filebrowser/debian/changelog
+++ b/deb/openmediavault-filebrowser/debian/changelog
@@ -1,6 +1,6 @@
 openmediavault-filebrowser (8.0.5-1) stable; urgency=medium
 
-  *
+  * Issue #2198: Prevent admin password reset on container restart.
 
  -- Volker Theile <volker.theile@openmediavault.org>  Thu, 07 May 2026 22:07:06 +0200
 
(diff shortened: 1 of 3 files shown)


votdev:
> Removed the auth part completely from the config file as it was redundant. Default user and pass are already admin:admin.

Do you have a reference to the documentation or source code which shows this behaviour? I thought I had read that the admin password is generated randomly when the database is initialized. The resulting problem is that it is extremely user-unfriendly to obtain this password.

jamfrade:
> > Removed the auth part completely from the config file as it was redundant. Default user and pass are already admin:admin.
> 
> Do you have a reference to the documentation or source code which shows this behaviour? I thought I had read that the admin password is generated randomly when the database is initialized. The resulting problem is that it is extremely user-unfriendly to obtain this password.

The docs do not make a explicit reference, however I have tested the behavior on my end.
I could also find the information in the source code.

https://github.com/gtsteffaniak/filebrowser/blob/main/backend%2Fcommon%2Fsettings%2Fauth.go
Lines 51-52

https://github.com/gtsteffaniak/filebrowser/blob/main/backend%2Fdatabase%2Fstorage%2Fstorage.go
Lines 91-92

https://github.com/gtsteffaniak/filebrowser/blob/main/backend%2Fcommon%2Fsettings%2Fconfig.go
Line 697

votdev:
Thx

PR #2201, opened 2026-05-08
feat(usermgmt): add optional per-user TOTP two-factor authentication

### LINKED ISSUE #2200: Add native per-user TOTP two-factor authentication (2FA)
**Is your feature request related to a problem? Please describe.**

  OpenMediaVault currently has no built-in second authentication factor.
  A stolen or guessed password is sufficient to gain full administrative
  access to the NAS. Users who want MFA today must deploy a reverse-proxy
  with an external identity provider (Authentik, Authelia, etc.), which
  adds significant infrastructure complexity for a home or small-office
  NAS.

  **Desired solution**

  Add optional, per-user TOTP (Time-based One-Time Password, RFC 6238)
  as a native second factor in the OMV login flow. After entering the
  correct password the user is prompted for a 6-digit code from any
  standard authenticator app (Google Authenticator, Authy, Microsoft
  Authenticator, 1Password, etc.).

  Key design decisions:

  - **Opt-in per account** — existing accounts are unaffected; each user
    enables it voluntarily from their profile page.
  - **Works for all accounts** — including the built-in `admin` account.
  - **No third-party services** — the QR code is generated entirely in
    the browser; the secret never leaves the server.
  - **Brute-force protection** — a pending MFA session is invalidated
    after 5 failed code attempts.
  - **Zero new dependencies** — the PHP implementation is pure RFC 6238
    with no Composer packages required.

  **Alternatives considered**

  - **External auth proxy (Authelia / Authentik)** — solves the problem
    but requires additional infrastructure outside OMV, which is
    impractical for most users (see #1747).
  - **FIDO2 / WebAuthn hardware keys** — stronger but requires hardware
    the majority of users do not have; could be a follow-up feature.
  - **Email OTP** — depends on a working mail configuration, which is
    not always set up on a NAS.

allanbarcelos:
Implement optional RFC 6238 time-based one-time passwords (TOTP) as second factor authentication

Implements RFC 6238 time-based one-time passwords as an optional second factor for all user accounts, including admin. Compatible with Google Authenticator, Authy, Microsoft Authenticator, 1Password, and other standard TOTP apps.

- Pure PHP TOTP implementation with no Composer dependencies
- Separate configuration model so MFA can be enabled per-user including admin accounts
- Provisional secret is only persisted after successful user verification
- Disabling MFA requires current valid TOTP code to prevent stolen-session abuse
- QR code generation is handled entirely in the browser; secret never leaves server-side
- Session enters `mfa_pending` state between login steps with 5-attempt brute-force limit before invalidation
- 489/489 unit tests passing

Fixes: https://github.com/openmediavault/openmediavault/issues/2200

Signed-off-by: Allan Barcelos <allan@barcelos.dev>

### DIFF
--- a/deb/openmediavault/usr/share/php/openmediavault/totp.php
+++ b/deb/openmediavault/usr/share/php/openmediavault/totp.php
@@ -1,0 +1,594 @@
+<?php
+
+/**
+ * This file is part of OpenMediaVault.
+ *
+ * @license   https://www.gnu.org/licenses/gpl.html GPL Version 3
+ * @author    Volker Theile <volker.theile@openmediavault.org>
+ * @copyright Copyright (c) 2009-2026 Volker Theile
+ *
+ * OpenMediaVault is free software: you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * OpenMediaVault is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with OpenMediaVault. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+namespace OMV;
+
+/**
+ * RFC 6238 Time-based One-Time Password (TOTP) implementation.
+ *
+ * Produces 6-digit codes on a 30-second window using HMAC-SHA1,
+ * compatible with Google Authenticator, Authy, Microsoft Authenticator,
+ * 1Password, and other standard authenticator apps.
+ *
+ * No external dependencies — pure PHP with standard functions only.
+ */
+class Totp
+{
+    /** Number of digits in each OTP code (RFC 6238 default). */
+    public const CODE_LENGTH = 6;
+
+    /** Duration of each time window in seconds (RFC 6238 default). */
+    public const TIME_STEP = 30;
+
+    /**
+     * How many adjacent windows to accept on each side of the current
+     * one. A value of 1 tolerates up to 30 seconds of clock skew
+     * between the authenticator app and the server.
+     */
+    public const ALLOWED_DRIFT = 1;
+
+    /**
+     * Generate a cryptographically secure 160-bit (20-byte) random
+     * secret encoded as an uppercase base32 string (RFC 4648, no
+     * padding).
+     *
+     * @return string 32-character uppercase base32-encoded secret.
+     */
+    public static function generateSecret(): string
+    {
+        return self::base32Encode(random_bytes(20));
+    }
+
+    /**
+     * Verify a 6-digit TOTP code submitted by the user.
+     *
+     * Accepts codes from ALLOWED_DRIFT windows before and after the
+     * current window to handle minor clock differences between the
+     * authenticator app and the server. Uses hash_equals() to prevent
+     * timing side-channel attacks.
+     *
+     * @param string $secret The base32-encoded TOTP secret for this user.
+     * @param string $code   The 6-digit code entered by the user.
+     * @return bool TRUE if the code is valid within the tolerance window.
+     */
+    public static function verifyCode(string $secret, string $code): bool
+    {
+        // Reject obviously malformed codes before doing any crypto work.
+        if (!preg_match('/^\d{6}$/', $code)) {
+            return false;
+        }
+        $now = (int) floor(time() / self::TIME_STEP);
+        for ($i = -self::ALLOWED_DRIFT; $i <= self::ALLOWED_DRIFT; $i++) {
+            if (hash_equals(self::computeCode($secret, $now + $i), $code)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Build the otpauth:// provisioning URI to enroll an authenticator app.
+     *
+     * The URI encodes the secret, issuer, and algorithm parameters. It
+     * should be rendered as a QR code on the frontend so the user can
+     * scan it with their authenticator app.
+     *
+     * @param string $secret   The base32-encoded secret.
+     * @param string $username The account name shown in the authenticator.
+     * @param string $issuer   The service name shown in the authenticator.
+     * @return string Full otpauth:// URI.
+     */
+    public static function getProvisioningUri(
+        string $secret,
+        string $username,
+        string $issuer
+    ): string {
+        $query = http_build_query([
+            'secret' => strtoupper($secret),
+            'issuer' => $issuer,
+            'algorithm' => 'SHA1',
+            'digits' => self::CODE_LENGTH,
+            'period' => self::TIME_STEP,
+        ]);
+        return sprintf(
+            'otpauth://totp/%s:%s?%s',
+            rawurlencode($issuer),
+            rawurlencode($username),
+            $query
+        );
+    }
+
+    /**
+     * Generate a QR code for the given URI and return it as a data URL.
+     *
+     * Produces an SVG QR code entirely in PHP with no external dependencies.
+     * The result is a base64-encoded data URL suitable for use directly as
+     * an HTML image source: data:image/svg+xml;base64,...
+     *
+     * Implements QR Code Model 2, error-correction level M, with automatic
+     * version selection. The algorithm follows ISO/IEC 18004:2015.
+     *
+     * @param string $data  The string to encode (typically an otpauth:// URI).
+     * @param int    $scale Pixel size of each module. Defaults to 4.
+     * @return string A data URL string (data:image/svg+xml;base64;...).
+     */
+    public static function generateQrDataUrl(string $data, int $scale = 4): string
+    {
+        $matrix = self::buildQrMatrix($data);
+        $size = count($matrix);
+        $quiet = 4; // quiet zone in modules
+        $total = ($size + $quiet * 2) * $scale;
+
+        $rects = '';
+        foreach ($matrix as $row => $cols) {
+            foreach ($cols as $col => $dark) {
+                if ($dark) {
+                    $x = ($col + $quiet) * $scale;
+                    $y = ($row + $quiet) * $scale;
+                    $rects .= sprintf(
+                        '<rect x="%d" y="%d" width="%d" height="%d"/>',
+                        $x,
+                        $y,
+                        $scale,
+                        $scale
+                    );
+                }
+            }
+        }
+        $svg = sprintf(
+            '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 %1$d %1$d"'
+            . ' width="%1$d" height="%1$d"><rect width="100%%" height="100%%"'
+            . ' fill="#fff"/><g fill="#000">%2$s</g></svg>',
+            $total,
+            $rects
+        );
+        return 'data:image/svg+xml;base64,' . base64_encode($svg);
+    }
+
+    /**
+     * Build a boolean matrix representing the dark/light modules of a QR code.
+     *
+     * Supports byte-mode encoding with error-correction level M. Selects the
+     * smallest QR version that can hold $data. Implements masking pattern 0
+     * (checkerboard: (row + col) % 2 == 0).
+     *
+     * @param string $data The input string to encode.
+     * @return array<int,array<int,bool>> 2-D array indexed [row][col].
+     */
+    private static function buildQrMatrix(string $data): array
+    {
+        // QR code capacity table (byte mode, error-correction level M).
+        // Index = version - 1; value = max bytes that version can hold.
+        static $capacities = [
+            16, 28, 44, 64, 86, 108, 124, 154, 182, 216,
+            254, 290, 334, 365, 415, 453, 507, 563, 627, 669,
+            714, 782, 860, 914, 1000, 1062, 1128, 1193, 1267, 1373,
+            1455, 1541, 1631, 1725, 1812, 1914, 1992, 2102, 2216, 2334,
+        ];
+        $byteLen = strlen($data);
+        $version = 1;
+        foreach ($capacities as $v => $cap) {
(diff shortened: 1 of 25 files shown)


deb/openmediavault/var/www/openmediavault/rpc/session.inc
```diff
@@ -33,76 +35,213 @@ class OMVRpcServiceSession extends \OMV\Rpc\ServiceAbstract {
...
+				OMV_ROLE_ADMINISTRATOR : OMV_ROLE_USER;
 			$session = &\OMV\Session::getInstance();
+			// Check whether this user has TOTP MFA enabled. If so, initialize
+			// a pending session and return early — the client must submit a
+			// TOTP code via verifyTotp() before receiving a full session.
+			if (\OMV\Totp::isEnabledForUser($params['username'])) {
+				if (!$session->isMfaPending()) {
+					$sessionId = $session->initializeMfaPending(
+						$params['username'], $role);
+				}
```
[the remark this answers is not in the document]
allanbarcelos: Fixed. Added a check for an existing authenticated session belonging to a different user before calling initializeMfaPending(), mirroring the same guard already present on the non-MFA path.

deb/openmediavault/var/www/openmediavault/rpc/session.inc
```diff
@@ -33,76 +35,213 @@ class OMVRpcServiceSession extends \OMV\Rpc\ServiceAbstract {
...
+				syslog(LOG_INFO, sprintf("MFA challenge issued to %s ".
+					"[username=%s, user-agent=%s]", $server->getClientIP(),
+					$params['username'], $server->getUserAgent()));
+				return [
+					"authenticated" => FALSE,
+					"mfaRequired"   => TRUE,
+					"username"       => $params['username'],
+					"permissions"    => $permissions,
+					"sessionid"      => $sessionId,
+				];
```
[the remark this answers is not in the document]
allanbarcelos: Fixed. Added an else branch that returns session_id() when the session is already in mfa_pending state, ensuring the client always receives a usable session ID to continue submitting TOTP codes.

deb/openmediavault/workbench/src/app/pages/usermgmt/users/user-mfa-form-page.component.ts
```diff
@@ -0,0 +1,193 @@
...
+      .subscribe({
+        next: (res: { secret: string; uri: string }) => {
+          this.provisionalSecret = res.secret;
+          // Generate the QR code image entirely in the browser so the
+          // secret never leaves the current page or reaches a third-party.
+          QRCode.toDataURL(res.uri, { errorCorrectionLevel: 'M', width: 256 })
+            .then((dataUrl: string) => {
+              this.qrCodeDataUrl = dataUrl;
+              this.setupCodeControl.reset();
+              this.state = 'setup';
```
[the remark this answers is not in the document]
allanbarcelos: Fixed. Added a .catch() handler to QRCode.toDataURL() that clears the provisional secret and surfaces an error notification to the user, preventing the component from getting stuck in an inconsistent state.

votdev:
@allanbarcelos I'll review it asap.

votdev:
The `UserMfaFormPageComponent` component shouldn't be somewhere under `pages/`; that directory is only meant for pages that follow the declarative approach. Ideally, all the pages there should be replaced with YAML files, but debugging is just so much easier this way.

After writing this i had a look into the MFA form. You NEED to convert it to declarative style. The way you implemented it is forbidden. Only core components, e.g. the Intuition module which is responsible for rendering the declarative code in the UI, are allowed to make use of Angular Material modules.

This is an example how a form MUST look like in OMV:
```
public config: FormPageConfig = {
    request: {
      service: 'UserMgmt',
      get: {
        method: 'getUser',
        params: {
          name: '{{ _routeParams.name }}'
        }
      },
      post: {
        method: 'setUser'
      }
    },
    fields: [
      {
        type: 'textInput',
        name: 'name',
        label: gettext('Name'),
        disabled: '{{ _routeConfig.data.editing | toboolean }}',
        value: '{{ _routeParams.name }}',
        autocomplete: 'off',
        validators: {
          required: true,
          patternType: 'userName'
        }
      },
      {
        type: 'textInput',
        name: 'email',
        label: gettext('Email'),
        value: '',
        autocomplete: 'off',
        validators: {
          patternType: 'email'
        }
      },
      {
        type: 'passwordInput',
        name: 'password',
        label: gettext('Password'),
        value: '',
        autocomplete: 'new-password',
        validators: {
          required: '{{ _routeConfig.data.editing | toboolean === false }}'
        }
      },
      {
        type: 'passwordInput',
        name: 'passwordconf',
        label: gettext('Confirm password'),
        submitValue: false,
        value: '',
        autocomplete: 'new-password',
        validators: {
          custom: [
            {
              constraint: {
                operator: 'eq',
                arg0: { prop: 'password' },
                arg1: { prop: 'passwordconf' }
              },
              errorData: gettext('Password does not match.')
            }
          ]
        }
      },
      {
        type: 'select',
        name: 'shell',
        label: gettext('Shell'),
        placeholder: gettext('Select a shell ...'),
        value: '/usr/bin/sh',
        store: {
          proxy: {
            service: 'System',
            get: {
              method: 'getShells'
            }
          },
          sorters: [
            {
              dir: 'asc',
              prop: 'text'
            }
          ]
        }
      },
      {
        type: 'select',
        name: 'groups',
        label: gettext('Groups'),
        placeholder: gettext('Select groups ...'),
        multiple: true,
        valueField: 'name',
        textField: 'name',
        value: [],
        store: {
          proxy: {
            service: 'UserMgmt',
            get: {
              method: 'enumerateAllGroups'
            }
          },
          sorters: [
            {
              dir: 'asc',
              prop: 'name'
            }
          ]
        }
      },
      {
        type: 'datatable',
        name: 'sshpubkeys',
        label: gettext('SSH public keys'),
        hasHeader: false,
        hasFooter: false,
        columns: [
          {
            prop: 'sshpubkey',
            flexGrow: 1,
            cellClass: 'omv-text-wrap omv-text-monospace'
          }
        ],
        actions: [
          {
            template: 'add',
            formDialogConfig: {
              title: gettext('SSH public key'),
              fields: [
                {
                  type: 'fileInput',
                  name: 'sshpubkey',
                  value: '',
                  rows: 10,
                  wrap: 'off',
                  monospace: true,
                  trim: true,
                  label: gettext('Public key'),
                  hint: gettext('The SSH public key in OpenSSH or RFC 4716 format.'),
                  validators: {
                    required: true,
                    patternType: 'sshPubKey'
                  }
                }
              ]
            }
          },
          {
            template: 'delete'
          }
        ],
        valueType: 'string',
        value: []
      },
      {
        type: 'checkbox',
        name: 'disallowusermod',
        label: gettext('Disallow account modification'),
        value: false,
        hint: gettext('Disallow the user to modify their own account.')
      },
      {
        type: 'tagInput',
        name: 'comment',
        label: gettext('Tags'),
        value: '',
        validators: {
          maxLength: 65
        }
      }
    ],
    buttons: [
      {
        template: 'submit',
        execute: {
          type: 'url',
          url: '/usermgmt/users'
        }
      },
      {
        template: 'cancel',
        execute: {
          type: 'url',
          url: '/usermgmt/users'
        }
      }
    ]
  };
```
Here are a few examples of what I think shouldn't be done ([example 1](https://github.com/ceph/ceph/blob/main/src/pybind/mgr/dashboard/frontend/src/app/ceph/cluster/services/service-form/service-form.component.html), [example 2](https://github.com/harvester/harvester-ui-extension/blob/main/pkg/harvester/edit/harvesterhci.io.virtualmachineimage.vue)). This is going to turn into a nightmare at some point.

If there isn't already a way to implement this, then the framework will need to be enhanced; e.g. by adding Material form components (see [here](https://github.com/openmediavault/openmediavault/tree/master/deb/openmediavault/workbench/src/app/core/components/intuition/material)). Of course, there's always the option to work around it a bit in exceptional cases if there is no other way, which is possible in this case since the form isn't part of a plugin.

An example is https://github.com/openmediavault/openmediavault/blob/master/deb/openmediavault/workbench/src/app/pages/storage/shared-folders/shared-folder-acl-form-page.component.ts. But even here, the form must follow the declarative approach.

You can find a small documentation at https://docs.openmediavault.org/en/8.x/development/plugins.html. Otherwise, unfortunately, the rule is: Use the source, Luke.

I know these are high standards, but that's the only way it works at OMV. From professional projects, I know what a nightmare it is when every page of an app implements everything on its own using Angular or Bootstrap. Eventually, that leads to a jumbled mess of code that becomes unmaintainable. That’s why OMV now exclusively uses the declarative approach for all pages and plugins. If something needs to be changed in the style of a form control, there’s exactly one place that needs to be modified.

votdev:
Can you please add some screenshots or videos of the login process and the UI that is responsible for configuring the MFA as user.

votdev:
I can not get it working.
grafik

deb/openmediavault/usr/share/php/openmediavault/totp.php
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	/**
+	 * Query the config database to check whether TOTP is enabled for a user.
+	 *
+	 * @param string $username The username to check.
+	 * @return bool TRUE if TOTP MFA is enabled and a secret is stored.
+	 */
+	public static function isEnabledForUser(string $username): bool {
+		$db = \OMV\Config\Database::getInstance();
```
votdev: There is an unwritten rule in OMV that database access should not be performed in classes or code located below the `php/openmediavault` directory. In fact, there is only one place where this is currently done, and that is in the RPCs or `omv-engined`.  The OMV PHP library must not have any knowledge of the database, except for the database class, of course. This is to prevent the code and logic from becoming mixed at the lowest level. Please find another solution so as not to break this rule.

deb/openmediavault/usr/share/php/openmediavault/totp.php
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	/**
+	 * Retrieve the stored TOTP secret for a user from the config database.
+	 *
+	 * @param string $username The username whose secret to retrieve.
+	 * @return string The base32-encoded secret, or '' if not set.
+	 */
+	public static function getSecretForUser(string $username): string {
+		$db = \OMV\Config\Database::getInstance();
```
votdev: Relocate this code into the RPC for example.

votdev (changes requested):

deb/openmediavault/workbench/src/app/pages/usermgmt/users/user-mfa-form-page.component.html
```diff
@@ -0,0 +1,139 @@
+<mat-card class="omv-mfa-card">
```
votdev: You need to implement that form in a declarative way.

deb/openmediavault/workbench/src/app/pages/usermgmt/usermgmt.module.ts
```diff
@@ -1,5 +1,6 @@
 import { CommonModule } from '@angular/common';
 import { NgModule } from '@angular/core';
+import { ReactiveFormsModule } from '@angular/forms';
```
votdev: It is not allowed to import such modules at that level.

votdev (changes requested):

allanbarcelos:
> Can you please add some screenshots or videos of the login process and the UI that is responsible for configuring the MFA as user.


https://github.com/user-attachments/assets/864687a2-6be1-4ab2-bc0c-e6674dc70770


https://github.com/user-attachments/assets/3c22b099-d85f-4f36-a40c-142be5021b51

allanbarcelos:
> scree

08-mfa-enabled
10-mfa-disabled-toast
09-disable-code
07-mfa-enabled-toast
06-code-entered
05-mfa-setup
04-mfa-disabled
03-dashboard
02-login-filled
01-login

votdev:
Thanks for the changes. I'll take a look at them as soon as I can. I'm a bit short on time right now, but the feature is definitely cool.

I'll probably have to do some of the work myself, which will take quite a bit of time. So first, I need to get through all the other items on my to-do list.

votdev:
I had to do some major cleanup first because the commits were all over the place.
But until the feature is integrated, I still have to make some significant changes. For example, the database migration is missing. I’m also not happy with the MFA configuration, though I have to admit that I don’t really understand it yet and have only watched the videos.

allanbarcelos:
> I had to do some major cleanup first because the commits were all over the place. But until the feature is integrated, I still have to make some significant changes. For example, the database migration is missing. I’m also not happy with the MFA configuration, though I have to admit that I don’t really understand it yet and have only watched the videos.

Hi @votdev, thanks for the cleanup and for your patience with this PR.
                                                                                                            
  I want to give you an update on where things stand after your feedback:                                 
                                                                                                            
  Already addressed:                                                                                        
  - The UserMfaFormPageComponent was rewritten using the declarative FormPageConfig approach — no Angular   
  Material modules are used directly in the component or in usermgmt.module.ts                              
  - A new form-image field type was added to the Intuition framework to display the QR code inside the    
  declarative form                                                                                          
  - All database access was removed from totp.php — it is now a pure PHP TOTP/QR implementation with no     
  knowledge of the config DB; all DB interaction lives in the RPCs                                     
  - The 3 bugs flagged by Copilot (cross-user session guard, empty session ID on MFA-pending state,         
  unhandled QR code promise rejection) were all fixed                                                     
                                                                                                            
  What I know is still missing:                                                                           
  - The database migration script (confdb/migrations.d/) and the <mfa> node in the config.xml template — I can 
   working on this                                                                                      
                                                                                                            
  Regarding your comment about not being fully happy with the MFA configuration flow: could you share what  
  specifically feels off? Whether it's the UX, the data model, or how the secret is stored — I'd rather     
  align with your vision now than have you do another round of cleanup.

votdev:
@allanbarcelos Please do not rebase the PR. I will take care of that. I need to implement several new features that this PR will then make use of.

votdev:
Thank you for the contribution. I've refactored it into a plugin. I had to replace the TOTP class because the generated QR code was not correct, no authenticator app that i used was able to read it.

Superseded by https://github.com/openmediavault/openmediavault/pull/2242.

PR #2203, opened 2026-05-12
fix(nginx): ensure security headers are sent for index.html and JSON responses

jeffcrouse:
## Problem

nginx silently drops server-level `add_header` directives for any location block that defines its own `add_header` directives ([nginx docs](https://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header)):

> These directives are inherited from the previous configuration level if and only if there are no `add_header` directives defined on the current level.

The `location ~* /index.html|\.json$` block in `site-webgui.j2` sets `Cache-Control` and `Pragma` via `add_header`. This causes **all** security headers defined in `security.conf` (CSP, HSTS, X-Frame-Options, X-Content-Type-Options, etc.) to be silently omitted from every response to `index.html` — the entry point of the Angular SPA.

The result: browsers never receive the security headers for the main page load, causing browsers like Chrome to display a "Not Secure" warning despite a valid TLS certificate.

This can be verified on an affected installation:

```
$ curl -si https://<omv-host>/ | grep -i "content-security\|x-frame\|strict-transport"
(no output)
```

## Fix

Split `security.conf.j2` into two files:

- **`security.conf.j2`** — TLS/SSL settings only (`ssl_protocols`, `ssl_ciphers`, etc.), which are only valid in the server context
- **`headers.conf.j2`** — HTTP security response headers (`add_header` directives), which can be included in both server and location contexts

Include `headers.conf` explicitly inside the `location ~* /index.html|\.json$` block so security headers are applied alongside the cache-control headers.

After this fix:

```
$ curl -si https://<omv-host>/ | grep -i "content-security\|x-frame\|strict-transport"
strict-transport-security: max-age=15768000; includeSubDomains
content-security-policy: default-src 'self'; script-src 'self' 'unsafe-eval' 'unsafe-inline'; ...
x-frame-options: SAMEORIGIN
```

## Files changed

- `10webgui.sls` — add Salt state to render `headers.conf` from `headers.conf.j2`
- `files/security.conf.j2` — keep only TLS/SSL settings
- `files/headers.conf.j2` — new file with HTTP security `add_header` directives (moved from `security.conf.j2`)
- `files/site-webgui.j2` — include `headers.conf` inside the `index.html|\.json$` location block


votdev:
The commit is missing your `Signed-off-by` line (see [here](https://github.com/openmediavault/openmediavault/blob/master/CONTRIBUTING.md)). Please add it, otherwise the DCO check will not allow to merge the PR. Thx

votdev:
Thanks for the hint and work. The issue has been fixed by https://github.com/openmediavault/openmediavault/commit/b727a351e56a90a1ea44a7418c8b6539dc1366b8.

PR #2226, opened 2026-06-06
restart monit if monit reload fails

ryecoaaron:
We have been seeing quite a few problems with monit failing to reload lately.  I have even seen this on my own systems.  monit is in a stale state and a reload does not clear the stale file(s) in /var/tmp.  Doing a monit restart fixes the issue in every case I have seen.

salt will try to reload monit unit and if it fails, it will restart the unit.


votdev (approved):
LGTM, thx

Issue #1554, opened 2023-05-17
Upgrade to Angular >= 15


votdev:
Migration has been done to v14 only till now because this has not required a huge code refactoring. With Angular Material >= 15 the UI code needs huge refactoring.

PR #2228, opened 2026-06-10
Upgrade Angular 19

### LINKED ISSUE #1783: Angular 14 EOL
Hi, would you consider a PR to update the angular version to the last v18?
Also with the last version there are some fancy news as standalone components, new flow syntax, vite bundler etccc

DDAZZA:
# Upgrade Angular 15 → 19

Performs a sequential major-version upgrade of the Angular workbench from v15 to v19.

Fixes: #1783 , #1554


votdev:
@DDAZZA Thanks for your work, but the PR can't be included in OMV8 because this is a stable release and I won't be making such extensive changes there. This can happen in OMV9 at the earliest.

DDAZZA:
@votdev Thanks for the clarification,

I do think there’s a good opportunity in the next release to improve how the UX/UI presents the functionality that’s already available.  From a new user’s perspective, a lot of the project’s strengths aren’t immediately obvious, and that may make it harder for people to appreciate what OMV offers.

PR #2229, opened 2026-06-10
Add BackgroundImageLayoutComponent and PublicLayoutComponent

votdev:
Related to: https://github.com/openmediavault/openmediavault/pull/2201

- [ ] References issue

### DIFF
--- a/deb/openmediavault/workbench/src/app/core/components/layouts/background-image-layout/background-image-layout.component.scss
+++ b/deb/openmediavault/workbench/src/app/core/components/layouts/background-image-layout/background-image-layout.component.scss
@@ -1,0 +1,31 @@
+// This file is part of OpenMediaVault.
+//
+// @license   https://www.gnu.org/licenses/gpl.html GPL Version 3
+// @author    Volker Theile <volker.theile@openmediavault.org>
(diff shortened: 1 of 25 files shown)


deb/openmediavault/workbench/src/app/core/components/layouts/background-image-layout/background-image-layout.component.scss
```diff
@@ -0,0 +1,31 @@
...
+  width: 100%;
+}
+
+.background {
+  background-position: center;
+  background-repeat: no-repeat;
+  background-size: cover;
+  inset: 0;
+  position: fixed;
+}
```
DDAZZA, line 31: It might be worth considering adding an overlay to darken the image and increase contrast with the login form.

```suggestion
}

&::before {
  background: rgba(0, 0, 0, 0.35);
  content: '';
  inset: 0;
  position: absolute;
}
```

votdev: That's not a good idea, because it would affect all the background images that can be configured by the openmeiavault-wbprefs plugin.

Issue #1612, opened 2023-10-10
OMV-NUT plugin status display


boelle:
would be nice to see if connected and status from the ups

and if commands can be sent to the ups include buttons for the generic ones

PrplHaz4:
Services -> UPS should give you the current NUT readings

boelle:
only the settings
![image](https://github.com/openmediavault/openmediavault/assets/3832239/f1f90341-7aec-4721-9679-a9020b042b6b)

PrplHaz4:
> only the settings ![image](https://user-images.githubusercontent.com/3832239/274163006-f1f90341-7aec-4721-9679-a9020b042b6b.png)

Sorry I'm still on OMV5 - try `Diagnostics -> Services -> UPS`

boelle:
yep, they are there, but still

i would still vote for that a noob would expect them (or a few of them) to be in the plugin so that they can see there is a connection to the ups

votdev:
> yep, they are there, but still
> 
> i would still vote for that a noob would expect them (or a few of them) to be in the plugin so that they can see there is a connection to the ups

You mean that you want to see them within the `Services | UPS` page?

OMV has a clear division of where to find which information in the UI. Diagnostics has nothing to do in the configuration area. Therefore there is a dedicated `Diagnostics` area where this information can be found.

boelle:
> them within the `Services | UPS` page

not all of them, just enough to see that there is a connection to the ups

dacmot:
What about a dashboard widget?

votdev:
> What about a dashboard widget?

The `Services` widget already displays the status of the NUT service.

boelle:
yes, but only running.... it does not tell if connected etc

votdev:
If you provide any necessary information how to get these information i'll have a look on it and try to implement a new widget. I have no UPS anymore, so i can not do any test.

boelle:
if you can get voltages etc then its connected

so if you can displate voltage under diagnostic then its connected

dacmot:
I presume the current UPS diagnostic in OMV is just printing the output from calling `upsc myups@localhost` or something similar.

Based on that information, I would at the least trim that to the following lines (if available):
```
battery.charge
battery.runtime
ups.load
ups.status
ups.test.result
```

If you're feeling inspired, you could turn some of these into colored bars/levels, and turn the runtime (I assume the number is in seconds) into an HH:MM:SS format.

[...]

PR #2234, opened 2026-06-17
Add UPS dashboard widgets (Network UPS Tools)

iSimonWeb:
## Summary
Adds two dashboard widgets to the Network UPS Tools plugin:
- **UPS Information** — a grid of device/battery details.
- **UPS Battery** — a gauge of battery charge with the low-battery shutdown threshold marked.

Backed by two new RPC methods, `Nut.getBatteryStats` (a parsed `battery.*` object for the gauge) and `Nut.getInformation` (curated `[{key, value}]` rows for the grid), which share private `upsc` fetch/parse helpers. `Nut.getStats` is unchanged for callers — it still returns the raw output used by the NUT diagnostics page.

Implements the status-display request in #1612. The command-button idea mentioned there is out of scope for this PR.

## Screenshots
OMV Dashboard - Widgets preview
OMV Diagnostics/Services/UPS - upsc text blob still working

## Testing
Manually verified on OMV 7.7.24-7 (Sandworm) against a real UPS: both widgets render after restarting `openmediavault-engined` and running `omv-mkworkbench dashboard`. No PHPUnit tests — dashboard widgets across the repo ship without them, and the RPC shells out to `upsc`.

- [x] References issue

### DIFF
--- a/deb/openmediavault-nut/usr/share/openmediavault/engined/rpc/nut.inc
+++ b/deb/openmediavault-nut/usr/share/openmediavault/engined/rpc/nut.inc
@@ -20,24 +20,24 @@
  * You should have received a copy of the GNU General Public License
  * along with OpenMediaVault. If not, see <https://www.gnu.org/licenses/>.
  */
+
+namespace Engined\Rpc;
+
+require_once("openmediavault/functions.inc");
+
 class OMVRpcServiceNetworkUPSTools extends \OMV\Rpc\ServiceAbstract
 {
-    /**
-     * Get the RPC service name.
-     */
     public function getName()
     {
         return "Nut";
     }
 
-    /**
-     * Initialize the RPC service.
-     */
     public function initialize()
     {
         $this->registerMethod("get");
         $this->registerMethod("set");
         $this->registerMethod("getStats");
+        $this->registerMethod("getVariables");
     }
 
     /**
@@ -94,25 +94,96 @@ class OMVRpcServiceNetworkUPSTools extends \OMV\Rpc\ServiceAbstract
         $this->validateMethodContext($context, [
             "role" => OMV_ROLE_ADMINISTRATOR
         ]);
-        // Is service enabled?
         $db = \OMV\Config\Database::getInstance();
         $object = $db->get("conf.service.nut");
         if (false === $object->get("enable")) {
-            $stats = gettext("Service disabled");
-        } else {
-            $mode = $object->get("mode");
-            $cmdArgs = [];
-            if ("netclient" === $mode) {
-                $remoteUps = $object->get("upsname")."@".$object->get("netclienthostname");
-                $cmdArgs[] = escapeshellarg($remoteUps);
-            } else {
-                $cmdArgs[] = escapeshellarg($object->get("upsname"));
-            }
-            $cmd = new \OMV\System\Process("upsc", $cmdArgs);
-            $cmd->setRedirect2to1();
-            $cmd->execute($output);
-            $stats = implode("\n", $output);
+            return gettext("Service disabled");
         }
-        return $stats;
+        return implode("\n", $this->getUpscOutput($object));
+    }
+
+    /**
+     * Get the UPS variables as a flat "key => value" object for the
+     * dashboard widgets. Both the battery gauge and the information grid
+     * consume this single method; the grid maps it to rows via the
+     * widget's "itemTemplates", the gauge reads the charge props.
+     *
+     * Variable names have their dots replaced by underscores so they can
+     * be referenced in widget templates (e.g. "battery.charge" becomes
+     * "battery_charge"). Numeric values are cast to numbers; duration
+     * variables (e.g. "battery.runtime") stay as raw seconds and are
+     * formatted in the widget template via the "duration" filter.
+     *
+     * @param params The method parameters.
+     * @param context The context of the caller.
+     * @return An associative array of UPS variables. "battery_charge" and
+     *   "battery_charge_low" are always present as numbers (the gauge
+     *   needs them), defaulting to 0 when the service is disabled or the
+     *   UPS does not report them.
+     */
+    public function getVariables($params, $context)
+    {
+        $this->validateMethodContext($context, [
+            "role" => OMV_ROLE_ADMINISTRATOR
+        ]);
+        // The gauge always needs these as numbers; default to 0 when the
+        // service is disabled or the UPS does not report them.
+        $result = ["battery_charge" => 0, "battery_charge_low" => 0];
+        $db = \OMV\Config\Database::getInstance();
+        $object = $db->get("conf.service.nut");
+        if (false === $object->get("enable")) {
+            return $result;
+        }
+        $vars = $this->parseVars($this->getUpscOutput($object));
+        foreach ($vars as $key => $value) {
+            $result[str_replace(".", "_", $key)] = is_numeric($value) ? $value + 0 : $value;
+        }
+        return $result;
+    }
+
+    /**
+     * Run "upsc" for the configured UPS and return its output lines.
+     * @param object The "conf.service.nut" configuration object.
+     * @return An array containing the lines of the "upsc" output.
+     */
+    private function getUpscOutput($object)
+    {
+        $cmdArgs = [];
+        if ("netclient" === $object->get("mode")) {
+            $remoteUps = $object->get("upsname")."@".$object->get("netclienthostname");
+            $cmdArgs[] = escapeshellarg($remoteUps);
+        } else {
+            $cmdArgs[] = escapeshellarg($object->get("upsname"));
+        }
+        $cmd = new \OMV\System\Process("upsc", $cmdArgs);
+        $cmd->setRedirect2to1();
+        $cmd->setRequireOutput();
+        $cmd->execute($output);
+        return $output;
+    }
+
+    /**
+     * Parse "upsc" output into a "key => value" map. Lines without a
+     * ": " separator (e.g. the "Init SSL" notice) and lines whose key
+     * is not a lowercase variable name (e.g. upsd "Error:" messages)
+     * are ignored.
+     * @param output The "upsc" output lines.
+     * @return An associative array mapping variable names to values.
+     */
+    private function parseVars($output)
+    {
+        $vars = [];
+        foreach ($output as $line) {
+            $parts = explode(": ", $line, 2);
+            if (2 !== count($parts)) {
+                continue;
+            }
+            $key = trim($parts[0]);
+            if (!preg_match('/^[a-z][a-z0-9._]*$/', $key)) {
+                continue;
+            }
+            $vars[$key] = trim($parts[1]);
+        }
+        return $vars;
     }
 }
--- a/deb/openmediavault-nut/usr/share/openmediavault/workbench/dashboard.d/nut-battery.yaml
+++ b/deb/openmediavault-nut/usr/share/openmediavault/workbench/dashboard.d/nut-battery.yaml
@@ -1,0 +1,30 @@
+version: "1.0"
+type: dashboard-widget
+data:
+  id: 16b7007d-1585-4920-8893-073af0e3356c
+  permissions:
+    role:
+      - admin
+  title: _("UPS Battery")
+  description: _("Battery charge with the low-battery shutdown threshold marked.")
+  type: chart
+  reloadPeriod: 30000
+  chart:
+    type: advanced-gauge
+    maxHeight: "150px"
+    maxWidth: "150px"
+    min: 0
+    max: 100
+    request:
+      service: Nut
+      method: getVariables
+    label:
+      formatter: template
+      formatterConfig: "{{ value | tofixed(0) }}%"
+    dataConfig:
+      - label: _("Charge")
+        prop: battery_charge
+        backgroundColor: "#4cd964"
+      - label: _("Low")
+        prop: battery_charge_low
+        backgroundColor: "#ff3b30"
--- a/deb/openmediavault-nut/debian/changelog
+++ b/deb/openmediavault-nut/debian/changelog
@@ -1,3 +1,9 @@
+openmediavault-nut (8.0.4-1) stable; urgency=medium
+
+  * Issue #1612: Add dashboard widgets.
+
+ -- Volker Theile <volker.theile@openmediavault.org>  Fri, 24 Jul 2026 10:34:28 +0200
+
 openmediavault-nut (8.0.3-1) stable; urgency=medium
 
   * Upgrade to SaltStack 3008.
--- a/deb/openmediavault-nut/debian/control
+++ b/deb/openmediavault-nut/debian/control
@@ -9,7 +9,7 @@ Homepage: https://www.openmediavault.org
 
 Package: openmediavault-nut
 Architecture: all
-Depends: openmediavault (>= 8.4.0), nut, udev
+Depends: openmediavault (>= 8.5.5), nut, udev
 Priority: optional
 Description: openmediavault Network UPS Tools (NUT) plugin
  Network UPS Tools (NUT) is a client/server monitoring system that
--- a/deb/openmediavault-nut/usr/share/openmediavault/workbench/dashboard.d/nut-info.yaml
+++ b/deb/openmediavault-nut/usr/share/openmediavault/workbench/dashboard.d/nut-info.yaml
@@ -1,0 +1,50 @@
+version: "1.0"
+type: dashboard-widget
+data:
+  id: c812d727-6579-4e03-b2c1-756d93786664
+  permissions:
+    role:
+      - admin
+  title: _("UPS Information")
+  description: _("UPS device and battery information.")
+  type: grid
+  reloadPeriod: 30000
+  grid:
+    class: "omv-grid-cols-3"
+    item:
+      title: "{{ key }}"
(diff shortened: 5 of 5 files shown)


iSimonWeb:
PR #2234 adds read-only dashboard widgets (UPS status + battery charge), which covers the "see if connected / status" part of this request. I'd like to follow up with the interactive part — viewing/editing writable UPS variables and running instant commands — but wanted to check direction before building, since it touches power-control behaviour.

Rough plan:
- **Editable variables**: enumerate via `upsrw <ups>` into a table (name, description, type, value); set via `upsrw -s VAR=VALUE -u <user> -p <pass> <ups>`, validated against each variable's type/constraints.
- **Instant commands** (the "send commands" ask): list via `upscmd -l`, run via `upscmd <ups> <cmd>`.
- **Credentials**:
  - *standalone* — the plugin already provisions a local upsd admin user with `actions = set` / `instcmds = ALL`; the backend (root) reads its password from `/etc/nut/upsd.users`.
  - *netclient* — reads (`upsc`, `upsrw`/`upscmd` listing) already work over the network, so both tabs populate in either mode. Writes/commands reuse the stored netclient credentials, but that account connects as `slave` (monitor-only) and the remote `upsd.users` is outside our control — so a write only succeeds if the remote admin granted that user `actions = set`/`instcmds`. I'd attempt it and, on `ERR ACCESS-DENIED`, show a clear message about what to grant remotely (optionally add an explicit admin user/password field for netclient).

A few questions before I start:
1. Is this something you'd want in the plugin?
2. Placement — I'd mirror Services → SMB/CIFS (Settings / Shares) and give Services → UPS tabs: **Settings** (existing service config), **Variables** (editable `upsrw` values) and **Commands** (instant `upscmd` actions). Does that structure work, or would you prefer a single combined tab?
3. Guard rails for destructive instant commands (e.g. `load.off`, `shutdown.return`) — confirmation dialog, allowlist, or leave commands out for now?
4. For netclient writes, does the "reuse netclient creds + clear error if the remote denies it" approach above match what you'd want, or would you rather gate writes to standalone mode for now?

Happy to implement whichever shape you prefer.

deb/openmediavault-nut/usr/share/openmediavault/workbench/dashboard.d/nut-battery.yaml
```diff
@@ -0,0 +1,32 @@
...
+    type: advanced-gauge
+    maxHeight: "150px"
+    maxWidth: "150px"
+    min: 0
+    max: 100
+    request:
+      service: Nut
+      method: getStats
+      params:
+        format: "json"
```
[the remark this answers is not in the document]
iSimonWeb: Fixed in 290589d4b — added a transform defaulting battery_charge/battery_charge_low to 0, so the gauge always gets numeric values when the service is disabled, the UPS is offline, or the driver doesn't expose battery.charge.low. Used | default(0) to match the transform idiom already used across the workbench widgets.

deb/openmediavault-nut/usr/share/openmediavault/engined/rpc/nut.inc
```diff
@@ -83,35 +83,86 @@ class OMVRpcServiceNetworkUPSTools extends \OMV\Rpc\ServiceAbstract
...
+        }
+        $format = (is_array($params) && array_key_exists("format", $params))
+            ? $params["format"] : "text";
         $db = \OMV\Config\Database::getInstance();
         $object = $db->get("conf.service.nut");
+        // Is service enabled?
         if (false === $object->get("enable")) {
-            $stats = gettext("Service disabled");
+            return in_array($format, ["json", "list"]) ? [] : gettext("Service disabled");
+        }
```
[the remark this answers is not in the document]
iSimonWeb: Fixed in 290589d4b. The disabled branch now returns new \stdClass() for format: "json", so it JSON-encodes to {} (object) as documented. list still returns [] (that format genuinely is an array of rows), and the default text path is unchanged.

deb/openmediavault-nut/usr/share/openmediavault/engined/rpc/nut.inc
```diff
@@ -83,35 +83,89 @@ class OMVRpcServiceNetworkUPSTools extends \OMV\Rpc\ServiceAbstract
...
+                }
+            }
+            return $rows;
+        }
+        // "json": dots -> underscores, numbers cast.
+        $stats = [];
+        foreach ($vars as $key => $value) {
+            $stats[str_replace(".", "_", $key)] = is_numeric($value) ? $value + 0 : $value;
         }
         return $stats;
```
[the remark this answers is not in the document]
iSimonWeb: Fixed in 1bb62a7 — the json format now returns new \stdClass() ({}) when no variables parse, matching the disabled-branch handling, so it's always an object as documented.

votdev:
@iSimonWeb I think you should introduce a new RPC endpoint. The `getStats` RPC is meant for something different than providing the information that is used by the widgets.

iSimonWeb:
Makes sense. I'll keep `getStats` returning the raw upsc output for the diagnostics page and move the widget data to its own endpoint. To avoid duplicating the fetch (especially the netclient `ups@host` handling), I'll factor the upsc call into a private helper shared by `getStats` and the new endpoint.

One question on shape: the battery gauge needs a parsed object `({ battery_charge, battery_charge_low, … })`, while the info widget is a grid that loads row-shaped data `([{ key, value }])`. Would you prefer two purpose-specific endpoints (one object, one list), or a single endpoint with the info widget reworked to consume the object?

Happy to name them however fits the plugin's conventions.

votdev:
> Makes sense. I'll keep `getStats` returning the raw upsc output for the diagnostics page and move the widget data to its own endpoint. To avoid duplicating the fetch (especially the netclient `ups@host` handling), I'll factor the upsc call into a private helper shared by `getStats` and the new endpoint.
> 
> One question on shape: the battery gauge needs a parsed object `({ battery_charge, battery_charge_low, … })`, while the info widget is a grid that loads row-shaped data `([{ key, value }])`. Would you prefer two purpose-specific endpoints (one object, one list), or a single endpoint with the info widget reworked to consume the object?
> 
> Happy to name them however fits the plugin's conventions.

I think 2 endpoints are the best in that situation. The following names should match to all other RPCs that are used somewhere in a plugin / core:

- getBatteryStats
- getInformation

iSimonWeb:
Thanks for the cleanup, @votdev — I kept your changes (the `Engined\Rpc` namespace, `functions.inc`, and the `is_assoc_array`/`array_value` helpers) and built the endpoint split on top of them.

As suggested, `getStats` is left to do one thing — return the raw `upsc` output for the diagnostics page — and the widgets now use two dedicated methods that share private fetch/parse helpers:

- **`getBatteryStats`** → a parsed `battery.*` object for the gauge
- **`getInformation`** → curated `[{key, value}]` rows for the grid

### Info grid layout
The grid (`omv-grid-cols-3`) is organised as identity → live state → battery → thresholds:

| | | |
|---|---|---|
| Manufacturer | Model | Battery type |
| Status | Input voltage | Load |
| Battery voltage | Runtime | Battery date |
| Battery low | Charge warning | Runtime low |

Runtimes render as `29m52s`, voltages/percentages carry their units, and rows for variables the UPS doesn't report are skipped.

### Battery gauge
I dropped the `transform` I'd previously added to `nut-battery.yaml` — routing values through the display-template engine made the gauge legend show `NaN`. Instead, `getBatteryStats` seeds `battery_charge`/`battery_charge_low` to `0`, so the RPC always returns numbers (covering a disabled service, an offline UPS, or a driver that doesn't report those vars). The guard now lives in the backend.

### Updated Dashboard screenshot
OMV Dashboard with new NUT widgets

deb/openmediavault-nut/usr/share/openmediavault/engined/rpc/nut.inc
```diff
@@ -94,25 +94,155 @@ class OMVRpcServiceNetworkUPSTools extends \OMV\Rpc\ServiceAbstract
...
+    {
+        $cmdArgs = [];
+        if ("netclient" === $object->get("mode")) {
+            $remoteUps = $object->get("upsname")."@".$object->get("netclienthostname");
+            $cmdArgs[] = escapeshellarg($remoteUps);
+        } else {
+            $cmdArgs[] = escapeshellarg($object->get("upsname"));
+        }
+        $cmd = new \OMV\System\Process("upsc", $cmdArgs);
+        $cmd->setRedirect2to1();
```
votdev, line 159: If we can asure that there is always output, then we should make use of `setRequireOutput`.
```suggestion
        $cmd->setRedirect2to1();
        $cmd->setRequireOutput();
```

iSimonWeb:
@votdev Thanks for the review!
Pushed an amend addressing both points:
- `blank_lines_before_namespace` -> added the missing blank line before `namespace Engined\Rpc;` so it satisfies `PSR-12`.
- `setRequireOutput` -> added `$cmd->setRequireOutput();` right after `setRedirect2to1()` in `getUpscOutput()`, as suggested. upsc always produces output when the UPS is reachable, so an empty result now raises a clear exception instead of propagating silently.

Heads-up on verification: I normally smoke-test on my NAS326, but it runs OMV7, whose core openmediavault predates `OMV\System\Process::setRequireOutput()` (added in core 8.2.13), so the RPC fatals there with "Call to undefined method". Since openmediavault-nut already Depends: openmediavault (>= 8.4.0), the dependency is satisfied on any OMV8 install; I just can't exercise this specific line on my box, so it leans on CI / your environment. I did confirm on the NAS that `upsc` exits non-zero on an unknown/unreachable UPS, so `Process::execute()` throws `ExecException` and the widgets degrade to the dashboard error card as expected.

I also kept the `["battery_charge" => 0, "battery_charge_low" => 0]` defaults in `getBatteryStats()` deliberately — they're complementary to `setRequireOutput()`, not redundant: they cover the service-disabled early return (which never reaches `getUpscOutput()`) and a reachable UPS that doesn't report those specific vars; the gauge needs both fields present as numbers. `setRequireOutput()` only guards genuinely-empty output.

One UX question while we're here: the Information widget (grid) supports a native empty message, but the Battery gauge has no built-in "no data" state, its only graceful fallback is the framework's error card when the RPC throws. Worth a small follow-up to give the gauge a proper empty/error placeholder? Happy to open a separate issue/PR if you agree.

deb/openmediavault-nut/usr/share/openmediavault/engined/rpc/nut.inc
iSimonWeb: Implemented on 30d4de1.

votdev:
@iSimonWeb Please check https://github.com/openmediavault/openmediavault/commit/c6a4e08b726ed532da32c763e94971de2e5397c1. I've added the new transformer function `itemTemplates` to the `DataStore` service. You can now implement only one RPC that delivers the key/value object with the UPS information, and the widget can use `itemTemplates` to convert the object into an array of object items that the grid widget can process. Check out the unit test cases https://github.com/openmediavault/openmediavault/blob/c6a4e08b726ed532da32c763e94971de2e5397c1/deb/openmediavault/workbench/src/app/shared/services/data-store.service.spec.ts#L175.

iSimonWeb:
Thanks for `c6a4e08`, @votdev. `itemTemplates` is exactly the right primitive. I'll refactor the NUT widgets to a single RPC that returns the flat key/value UPS object: the **Information** grid maps it via `itemTemplates`, and the **Battery** gauge reads its props from the same response. The unit tests made the contract clear (output order follows the template array, fields pulled by name), really clean.

While wiring it up I hit two small gaps. Both are optional extensions to your new feature, only raising them since you opened the door to it.

### 1. A `duration` filter

The existing filters cover almost everything I need. `localedate` already handles `date` / `time` / `datetime` / `relative`, `notavailable` handles empty values, and `binaryunit` / `tofixed` cover the rest. The one thing missing is formatting a seconds count as a human duration. NUT reports it for `battery.runtime` (e.g. `1792`), and the same filter would suit an uptime value (days and hours), so I'd make it scale across d/h/m/s rather than just m/s:

```
{{ battery_runtime | duration }}         => 29m52s
{{ uptime | duration }}                  => 1d1m1s
{{ uptime | duration('long') }}          => 1 day 1 min 1 sec
```

Rough sketch:

```ts
nunjucksEnv.addFilter('duration', (value, style = 'short') => {
  let seconds = _.toInteger(value);
  if (!_.isFinite(seconds) || seconds < 0) { return value; }
  const units = [
    { short: 'd', long: 'day',  secs: 86400 },
    { short: 'h', long: 'hour', secs: 3600 },
    { short: 'm', long: 'min',  secs: 60 },
    { short: 's', long: 'sec',  secs: 1 }
  ];
  const parts = [];
  _.forEach(units, (u) => {
    const v = Math.trunc(seconds / u.secs);
    seconds %= u.secs;
    if (v > 0) {
      parts.push(style === 'long' ? `${v} ${u.long}${v !== 1 ? 's' : ''}` : `${v}${u.short}`);
    }
  });
  if (_.isEmpty(parts)) {
    parts.push(style === 'long' ? '0 secs' : '0s');
  }
  return parts.join(style === 'long' ? ' ' : '');
});
```

(The exact zero-unit trimming and unit labels are your call, this is just to show the shape.)

### 2. An optional `hidden` per `itemTemplates` entry

Since `itemTemplates` emits one item per template, I lose the "skip rows the UPS doesn't report" behaviour I have now (an absent var would render an empty row). A per-entry `hidden` template, evaluated during the map and skipped when truthy, would solve it, and it stays consistent with the `hidden?:` convention already used in form fields, datatable columns, and navigation:

```yaml
itemTemplates:
  - key: _("Runtime")
    value: "{{ battery_runtime | duration }}"
    hidden: "{{ not battery_runtime }}"      # drop the row if the var is absent
```

Because it's a full template against the whole response (not just the current field), it generalises nicely: hide a row based on a different field's presence, or any condition. Rough sketch in your loop:

```ts
_.forEach(store.itemTemplates, (template) => {
  if (!_.isUndefined(template.hidden) && toBoolean(formatDeep(template.hidden, store.data))) {
    return; // skip this row
  }
  mappedData.push(formatDeep(_.omit(template, ['hidden']), store.data));
});
```

(A `hideIfEmpty: '{{ field }}'` shorthand could sit on top of it, but `hidden` alone is the more flexible primitive.)

Both are independent of the NUT PR. Happy to open a separate PR for either if you'd prefer them tracked on their own, or to adjust the approach to your taste. If you'd rather keep the feature minimal, I can also preformat the duration server-side and filter absent rows in the RPC, since `itemTemplates` works great either way.

votdev:
Are you still working on the PR?

iSimonWeb:
Sorry for the late reply, yes I'm still on it!

I think my last comment may have slipped through, mind taking a quick look back? I ran into two small gaps with `itemTemplates` and proposed a `duration` filter and an optional `hidden` property to cover them.

I've already got the code ready locally with plug'n'play methods for both, so I'm happy either way: we can ship the current state as is (durations preformatted server side, `notavailable` for missing values), or extend `itemTemplates` with those two additions.

Just let me know what you prefer and I'll push.
I'll rebase on latest master meanwhile.

iSimonWeb:
Hey @votdev, I've pushed the update I mentioned. It's the `itemTemplates`-based version you suggested, and I kept your `>=8.5.0` bump. Ready for your review whenever you get a chance.

deb/openmediavault-nut/usr/share/openmediavault/engined/rpc/nut.inc
```diff
@@ -94,25 +94,122 @@ class OMVRpcServiceNetworkUPSTools extends \OMV\Rpc\ServiceAbstract
...
+        return $result;
+    }
+
+    /**
+     * Format a number of seconds as a "<minutes>m<seconds>s" string,
+     * e.g. 1792 becomes "29m52s". Non-numeric input is returned as is.
+     * @param value The number of seconds to format.
+     * @return The formatted duration string.
+     */
+    private function formatDuration($value)
```
votdev: You can now make use of the `duration` filter. See  https://github.com/openmediavault/openmediavault/pull/2247. Available with `openmediavault 8.5.2`.

iSimonWeb: Awesome, I will remove the RPC based duration format method to use the new filter and bump the OMV required version to `8.5.2` as you stated.
It'll be done next week since I'm on a brief holiday and my laptop's home, resting peacefully.
Cheers.

votdev: Please also check https://github.com/openmediavault/openmediavault/commit/17653a0ddcee256c16df31c3285611c02db03198. This format string is equal to the PHP code.

iSimonWeb: Fixed in 9411fde.

iSimonWeb:
Thanks @votdev! I've adopted the `duration` filter with the `mmm[m]ss[s]` format (matching the old PHP), dropped the server-side formatter, and bumped the dependency to `>= 8.5.2`.
Ready for review whenever you have a moment.

votdev (approved):
LGTM, thx

iSimonWeb:
Thanks @votdev! Really glad I got to collaborate on such a massive project. Appreciate the time you put into the `itemTemplates` addition that made everything even cleaner.

Cheers!

PR #2243, opened 2026-07-07
Issue #2106: Do not offer non-empty disks as file system candidates

### LINKED ISSUE #2106: being given the option to create a filesystem on a disk that already has one.
I have a zfs pool and when I go to create an xfs partition it is showing the zfs pools as being available to create a partition on.

I understand that zfs is a plugin but even then if a disk has partitions or a recognised filesystem (i.e. it hasn't been wiped) then imo it shouldn't show here.

I hope this makes sense, and you agree.
Thanks

Image

outhud:
Fixes #2106

## Problem
The **Create file system** device selection is populated by `FileSystemMgmt::getCandidates`. It only excluded devices used as slaves (LVM/mdraid) or whose file system is reported by `Filesystem::getFilesystems()`. A disk carrying a signature that OMV core does not enumerate — most notably **ZFS pool members**, but also foreign/unmounted file systems or a leftover partition table — was still offered as a candidate, allowing an accidental `mkfs` over data (e.g. a live ZFS pool, as reported in #2106).

## Change
- Add `StorageDevice::hasSignature()`, a `wipefs`-based probe (read-only, `--no*` list mode — it does **not** modify the device) that returns `TRUE` when a device carries any signature: a partition table, or a file system / RAID / LVM / LUKS member signature. This is the symmetric counterpart of the existing **Wipe** action — a device has a signature iff it needs wiping — and in one call also covers `crypto_LUKS`, `LVM2_member` and `linux_raid_member`.
- `getCandidates` now skips a device when `hasSignature()` returns `TRUE`. A truly empty/wiped disk returns no signature and remains a valid candidate for all file system types.
- Update the create-file-system forms (generic + btrfs) so the page hint and the device-select field hint explain that only empty devices are listed and that a device must be wiped to be reused.

## Notes / testing
I don't have OMV hardware to exercise this against real disks, so this has **not** been run end-to-end — I'd appreciate a maintainer smoke test:
1. With a `zfs_member` / LUKS / LVM / leftover-GPT disk present, confirm it no longer appears under Storage ▸ File Systems ▸ Create.
2. Wipe a disk and confirm it reappears as a candidate for ext4/xfs/btrfs.

The one runtime assumption is that `wipefs --parsable --noheadings --output type <dev>` prints one line per signature and nothing for an empty disk (standard util-linux behaviour). PHP files pass `php -l`; the two YAML component files parse cleanly.

### DIFF
--- a/deb/openmediavault/usr/share/openmediavault/workbench/component.d/omv-storage-filesystem-btrfs-create-form-page.yaml
+++ b/deb/openmediavault/usr/share/openmediavault/workbench/component.d/omv-storage-filesystem-btrfs-create-form-page.yaml
@@ -36,7 +36,7 @@ data:
         name: devicefiles
         label: _("Devices")
         placeholder: _("Select devices ...")
-        hint: _("Select the devices that will be used to create the file system. Only empty devices are listed; wipe a device to reuse it.")
+        hint: _("Select the devices that will be used to create the file system.")
         multiple: true
         valueField: devicefile
         textField: description
--- a/deb/openmediavault/usr/share/openmediavault/workbench/component.d/omv-storage-filesystem-btrfs-create-form-page.yaml
+++ b/deb/openmediavault/usr/share/openmediavault/workbench/component.d/omv-storage-filesystem-btrfs-create-form-page.yaml
@@ -6,7 +6,7 @@ data:
   config:
     hints:
       - type: info
-        text: _("If a device is not listed here, it is usually because the device already contains a file system or partitions. With the former, the file system can be mounted <a href='#/storage/filesystems/mount'>here</a>. For the latter, please <a href='#/storage/disks'>wipe</a> the device as partitions are not supported.")
+        text: _("If a device is not listed here, it is usually because the device is not empty, i.e. it already contains a file system, partitions or another signature such as a ZFS, LVM, Linux RAID or LUKS member. An existing file system can be mounted <a href='#/storage/filesystems/mount'>here</a>. Otherwise, please <a href='#/storage/disks'>wipe</a> the device to be able to reuse it.")
         dismissible: true
         stateId: 199e8c7a-ee3b-11ed-9fb1-2750842abc68
     fields:
@@ -36,7 +36,7 @@ data:
         name: devicefiles
         label: _("Devices")
         placeholder: _("Select devices ...")
-        hint: _("Select the devices that will be used to create the file system.")
+        hint: _("Select the devices that will be used to create the file system. Only empty devices are listed; wipe a device to reuse it.")
         multiple: true
         valueField: devicefile
         textField: description
--- a/deb/openmediavault/usr/share/openmediavault/workbench/component.d/omv-storage-filesystem-generic-create-form-page.yaml
+++ b/deb/openmediavault/usr/share/openmediavault/workbench/component.d/omv-storage-filesystem-generic-create-form-page.yaml
@@ -6,7 +6,7 @@ data:
   config:
     hints:
       - type: info
-        text: _("If a device is not listed here, it is usually because the device already contains a file system or partitions. With the former, the file system can be mounted <a href='#/storage/filesystems/mount'>here</a>. For the latter, please <a href='#/storage/disks'>wipe</a> the device as partitions are not supported.")
+        text: _("If a device is not listed here, it is usually because the device is not empty, i.e. it already contains a file system, partitions or another signature such as a ZFS, LVM, Linux RAID or LUKS member. An existing file system can be mounted <a href='#/storage/filesystems/mount'>here</a>. Otherwise, please <a href='#/storage/disks'>wipe</a> the device to be able to reuse it.")
         dismissible: true
         stateId: 199e8c7a-ee3b-11ed-9fb1-2750842abc68
     fields:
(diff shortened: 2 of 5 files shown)


votdev:
Thanks for the contribution. Please read the CONTRIBUTING.md how to sign commits correctly. Otherwise the DCO check will fail and the PR can’t be accepted.

ryecoaaron:
While I think this change is still worth adding, I cannot replicate your findings with zfs.  Are you on OMV 8 with openmediavault-zfs 8.1.1 installed?  The zfs plugin should not be enumerating zfs members for the filesystem create dialog.

outhud:
Thanks for testing this. You're right that on OMV 8 with openmediavault-zfs 8.1.1 the plugin filters out zfs pool members, so the pure "zfs plugin installed" path won't reproduce it — I should have framed #2106 better.

The point of this change is that it lives in **core** and is signature-agnostic, so it's defense-in-depth for the cases a plugin doesn't cover:
- a ZFS pool created/imported from the CLI (or a disk moved from another box) when the plugin isn't managing it
- **LUKS, LVM and foreign RAID members**, which nothing in core currently filters out of the create dialog

To confirm that last part I ran it on a clean OMV 8.1.1 VM. Stock `FileSystemMgmt.getCandidates` offered all three of an empty disk, a `crypto_LUKS` disk and an `LVM2_member` disk. With this patch, only the empty disk is offered — the LUKS and LVM disks are correctly excluded, and a wiped disk still shows up as a candidate for all fs types. So even setting ZFS aside, core is currently offering encrypted/LVM disks for `mkfs`, which is the same accidental-overwrite footgun as #2106.

outhud:
Thanks — I've rebased to a single commit with a proper `Signed-off-by` and added the `Fixes:` line from CONTRIBUTING.md. DCO is green now.

ryecoaaron:
> The point of this change is that it lives in core and is signature-agnostic, so it's defense-in-depth for the cases a plugin doesn't cover:

You don't have to justify the change.  I already agree with it.  I was just making sure you weren't seeing the issue with zfs since you bolded zfs members in the post.  I spent hundreds of hours on the rewrite and making sure it didn't show zfs members in the dropdown was something I spent a lot of time on.

> a ZFS pool created/imported from the CLI (or a disk moved from another box) when the plugin isn't managing it

The plugin used to auto-discover.  People didn't want that.  So, a manual discover would still add a pool created from the CLI to the list so they wouldn't show up in the creation dropdown.  Making large amounts of of changes to OMV to handle things done from the command line is a bad idea though.  The point of OMV is to use the web interface.  With the changes to the zfs plugin, there is no reason not to use the plugin or at least use the Discover feature.

outhud:
> > The point of this change is that it lives in core and is signature-agnostic, so it's defense-in-depth for the cases a plugin doesn't cover:
> 
> You don't have to justify the change. I already agree with it. I was just making sure you weren't seeing the issue with zfs since you bolded zfs members in the post. I spent hundreds of hours on the rewrite and making sure it didn't show zfs members in the dropdown was something I spent a lot of time on.
> 
> > a ZFS pool created/imported from the CLI (or a disk moved from another box) when the plugin isn't managing it
> 
> The plugin used to auto-discover. People didn't want that. So, a manual discover would still add a pool created from the CLI to the list so they wouldn't show up in the creation dropdown. Making large amounts of of changes to OMV to handle things done from the command line is a bad idea though. The point of OMV is to use the web interface. With the changes to the zfs plugin, there is no reason not to use the plugin or at least use the Discover feature.

Understood, yes that's right on the bolding - I didn't mean to imply the plugin was leaking members there, my mistake.  Thanks for the context on Discover, and for checking this.

votdev:
Need to check if calling another CLI command on every request is not too expensive when many disks are processed.

ryecoaaron:
@votdev on my dev vm with lots of LVM volumes, zfs volumes, and virtual disks, the change didn't seem to add much time.  On a slow system with lots of disks, it might be noticeable but I don't think unreasonable.

deb/openmediavault/usr/share/openmediavault/workbench/component.d/omv-storage-filesystem-btrfs-create-form-page.yaml
```diff
@@ -36,7 +36,7 @@ data:
         name: devicefiles
         label: _("Devices")
         placeholder: _("Select devices ...")
-        hint: _("Select the devices that will be used to create the file system.")
+        hint: _("Select the devices that will be used to create the file system. Only empty devices are listed.")
```
votdev: This is redundant. The hint `If a device is not listed here, it is usually because the device is not empty` already explains this.
If a text is changed, ALL previous translations become invalid. Therefore, changes should only be made if they make sense and if the risk of translations being lost and not retranslated is justified.

deb/openmediavault/usr/share/openmediavault/workbench/component.d/omv-storage-filesystem-generic-create-form-page.yaml
```diff
@@ -20,7 +20,7 @@ data:
         name: devicefile
         label: _("Device")
         placeholder: _("Select a device ...")
-        hint: _("Select the device that will be used to create the file system.")
+        hint: _("Select the device that will be used to create the file system. Only empty devices are listed.")
```
votdev: This is redundant. The hint `If a device is not listed here, it is usually because the device is not empty` already explains this.
If a text is changed, ALL previous translations become invalid. Therefore, changes should only be made if they make sense and if the risk of translations being lost and not retranslated is justified.

PR #2244, opened 2026-07-07
Issue #1796: Add page size selector to datatable footer

### LINKED ISSUE #1796: [feature reqest] Create a user or installation setting for number of items per page.

outhud:
Allow the user to change the number of items shown per page directly from the datatable's pagination bar.

An **Items per page** selector (10 / 25 / 50 / 100 / All) is added to the datatable footer, next to the pager. The chosen value is persisted as a global per-user setting through the existing `UserLocalStorageService`, so it is stored on the server (per user and device type) and restored on login — consistent with how column visibility and sorting are already handled. There is no new service and no backend change; the `WebGui.setLocalStorageItem` RPC already exists.

### Notes
- **Placement & persistence** follow the feedback given on the earlier #1900 (the selector lives in the pagination bar, and the value is stored via the same backend-synced mechanism as the other table settings). #1900 can be closed in favour of this.
- **Global on purpose:** the preference is intentionally global rather than per-table (per `stateId`), so setting it once applies everywhere. Selecting **All** is an explicit opt-in.
- Views that intentionally disable paging (`limit = 0`) are left untouched, and for remote-paged tables **All** is sent to the backend as `limit = -1` (the RPC treats any `limit >= 0` as a slice, so `0` would return no rows).

### Testing
- `npm run lint` (eslint + angular template rules + tsc + stylelint + prettier) — passes
- `ng test` for `datatable.component.spec.ts` — 8/8 pass (adds coverage for persistence, restore-on-init, the remote `All -> -1` path, and the paging-disabled guard)
- `ng build --configuration production` — clean AOT build

Fixes #1796

- [x] References issue
- [x] Includes tests for new functionality

### DIFF
--- a/deb/openmediavault/workbench/src/app/shared/components/datatable/datatable.component.html
+++ b/deb/openmediavault/workbench/src/app/shared/components/datatable/datatable.component.html
@@ -66,6 +66,48 @@
                [sortType]="sortType"
                (sort)="onSort($event)"
                (page)="onPage($event)">
+  <ngx-datatable-footer>
+    <ng-template ngx-datatable-footer-template
+                 let-rowCount="rowCount"
+                 let-pageSize="pageSize"
+                 let-selectedCount="selectedCount"
+                 let-curPage="curPage"
+                 let-offset="offset">
+      <div class="page-count">
+        <span *ngIf="selectionType !== 'none'">
+          {{ selectedCount?.toLocaleString() }} {{ messages.selectedMessage }} /
+        </span>
+        {{ rowCount?.toLocaleString() }} {{ messages.totalMessage }}
+      </div>
+      <div class="omv-datatable-footer-actions">
+        <mat-form-field *ngIf="pagingEnabled"
+                        class="omv-datatable-page-size">
+          <mat-select [value]="limit"
+                      (selectionChange)="onPageSizeChange($event.value)"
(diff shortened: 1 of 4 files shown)


ryecoaaron:
Looks good when there are more than one page
image
But when there are no pages, it needs some padding or margin to move it away from the datatable edge
image

PR #2270, opened 2026-09-08
fix(workbench): fix typo in detectSystemTheme() media query

Dr4w:
'prefers-color-schema' was used instead of 'prefers-color-scheme'
in the detectSystemTheme() method of PrefersColorSchemeService.

This typo prevented the automatic system theme detection from ever
working, causing OMV to always fall back to the default light theme
instead of respecting the user's OS-level dark/light preference.

Signed-off-by: Dr4w <dev.armando.local@gmail.com>


votdev (approved):
LGTM, thx.

